PR #8, opened 2026-01-12
Validate GitHub token permissions, add missing webhook events

d-mo:
> [!NOTE]
> - **Tracker registration (API/UI):** Validates GitHub token scopes and org admin access; returns `warnings` via new `TrackerCreateResponse` and surfaces them in the console without auto-closing the modal. Updates OpenAPI accordingly.
> - **GitHub/GitLab integration:** Adds `tracker_type` to base/subclasses; GitHub gains `validate_token_permissions` and registers additional PR-related webhook events. Normalizer maps new PR actions (review requested, synchronize, ready for review) and captures the specific requested reviewer.
> - **MCP endpoints:** More permissive PR/MR slug parsing and switch to `crud_organization.get_for_tracker` in multiple calls.
> - **Flows:** Cloned presets now start with `is_enabled=true`; trigger service logs clearer messaging for disabled flows; UI shows AI model, tracker/org/project names, optional prompt template, and better error handling.
> - **Build/runtime:** CI and docker-compose retag images to `ghcr.io/preloop/*` and add a `console` service.
> - **Tests:** New unit tests cover GitHub token permission validation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bbf2fe10da8eb43e4f12ceb6e81d349731ca9e10. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,7 +198,7 @@ jobs:
         uses: docker/metadata-action@v5
         with:
           images: |
-            ${{ env.REGISTRY_GHCR }}/${{ env.IMAGE_NAME }}
+            ${{ env.REGISTRY_GHCR }}/preloop/preloop
             ${{ vars.PUSH_TO_DOCKERHUB == 'true' && format('{0}/preloop', vars.DOCKERHUB_ORG) || '' }}
           tags: |
             type=ref,event=branch
@@ -224,7 +224,7 @@ jobs:
         uses: docker/metadata-action@v5
         with:
           images: |
-            ${{ env.REGISTRY_GHCR }}/${{ env.IMAGE_NAME }}/console
+            ${{ env.REGISTRY_GHCR }}/preloop/console
             ${{ vars.PUSH_TO_DOCKERHUB == 'true' && format('{0}/console', vars.DOCKERHUB_ORG) || '' }}
           tags: |
             type=ref,event=branch
--- a/backend/preloop/api/endpoints/flows.py
+++ b/backend/preloop/api/endpoints/flows.py
@@ -110,6 +110,7 @@ def clone_preset(
         raise HTTPException(status_code=404, detail="Preset not found")
 
     # Build dict excluding fields we want to override or that aren't in FlowCreate
+    # Note: is_enabled is excluded so cloned flows start enabled (presets are disabled)
     preset_dict = {
         k: v
         for k, v in preset.__dict__.items()
@@ -121,6 +122,7 @@ def clone_preset(
             "updated_at",
             "name",
             "is_preset",
+            "is_enabled",
             "account_id",
         ]
     }
@@ -140,6 +142,7 @@ def clone_preset(
         **preset_dict,
         name=final_name,
         is_preset=False,
+        is_enabled=True,  # Cloned flows start enabled
         account_id=str(current_user.account_id),
     )
     cloned_flow = crud_flow.create(
--- a/backend/preloop/api/endpoints/mcp.py
+++ b/backend/preloop/api/endpoints/mcp.py
@@ -935,8 +935,8 @@ async def add_comment(target: str, comment: str) -> "AddCommentResponse":
                 # Remove gitlab host and get project path (everything between host and /-/)
                 project_path = "/".join(url_path[1:]).rstrip("/-")
                 logger.info(f"Detected GitLab MR: {project_path}#{pr_mr_number}")
-    # Parse slug format for PRs/MRs: owner/repo#123
-    elif "/" in target and "#" in target:
+    # Parse slug format for PRs/MRs: owner/repo#123 or repo#123
+    elif "#" in target:
         slug_parts = target.split("#")
         pr_mr_number = slug_parts[1]
         project_path = slug_parts[0]
@@ -1156,7 +1156,7 @@ async def get_pull_request(pull_request: str) -> "PullRequestResponse":
         tracker = trackers[0]
         from preloop.models.crud import crud_organization
 
-        organizations = crud_organization.get_multi_by_tracker(
+        organizations = crud_organization.get_for_tracker(
             db, tracker_id=tracker.id, account_id=current_user.account_id
         )
         if not organizations:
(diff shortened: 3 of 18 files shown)


preloop-agent:
backend/preloop/api/endpoints/trackers.py:153
You’re only passing the first `ORGANIZATION` identifier into `validate_token_permissions`. If the tracker is scoped to multiple orgs (quite common for enterprise setups), we’ll silently skip the rest and never warn the user that their token lacks admin access or required scopes for those other orgs. That’s exactly the scenario this validation is trying to catch, so we end up giving a clean bill of health while downstream webhook registration still fails. Please iterate over every included organization (or at least collect warnings for each) before returning success.

preloop-agent:
backend/preloop/api/endpoints/trackers.py:59
The new response shape no longer fits the `Dict[str, str]` return annotation – we now emit a `warnings: List[str]`. FastAPI uses the type hint to generate OpenAPI, so the docs and any generated clients will tell callers that all values are strings. That makes the `warnings` payload effectively invisible to SDK users and can even break strict clients. Please widen the return signature (or add a proper response model) so the schema matches what we actually send back.

preloop-agent:
Thanks for adding the token validation flow! One thing I noticed is that the org-membership probe is using our stored scope identifier verbatim when building `/orgs/{identifier}/memberships/{username}`. In the UI we persist GitHub organization scope rules with the numeric org id that comes back from `get_organizations` (that method returns `id: str(org["id"])`, and `frontend/src/components/add-tracker-modal.ts:486` posts `identifier: org.id`). Passing a numeric id to `/orgs/{...}` yields a 404 from GitHub, so this new path will always emit the "You are not a member" warning even when the token is correct. Could we translate numeric identifiers to the org login (e.g. call `/organizations/{id}` first) or otherwise ensure we store the slug so the membership check hits `/orgs/{slug}/memberships/...`?

preloop-agent:
I spotted a couple of issues that will trip users up:

- **backend/preloop/sync/trackers/github.py:417** – the token permission check calls `GET /orgs/{org_identifier}/memberships/{username}` using the scope-rule identifier. Our GitHub sync stores organization identifiers as the numeric GitHub org ID (`transform_organization` sets `identifier = str(org_data["id"])`). Passing that numeric ID to the `/orgs/{slug}/memberships` endpoint always returns 404, so even valid tokens will come back with the "You are not a member" warning. We need to translate the stored identifier back to the org login (or look it up first) before probing membership; otherwise we’ll get false negatives everywhere and still miss real permission problems.
- **backend/preloop/api/endpoints/trackers.py:153** – when a tracker is scoped to multiple organizations, we only validate the first one. If the PAT lacks admin rights on the second/third org, registration still succeeds with no warning, and sync/webhook setup will later fail. We should iterate over every included org (aggregating errors/warnings per org) instead of short-circuiting after the first entry.

Once those two are addressed I didn’t see other blockers, but these are both worth fixing before merge.

preloop-agent:
**Issue (blocking):** When `org_identifier` is numeric we resolve it via `/organizations/{id}`. If that lookup fails (403/404 is common when the token is missing `read:org` or is a fine-grained PAT), `org_login` is left as the numeric string. We then fall through to the membership check and call `GET /orgs/{org_login}/memberships/{username}`. GitHub expects the org *slug* there, so the call returns 404 and we emit the "You are not a member of organization '123456'" warning. In reality the token may be fine – we just could not resolve the ID. This ends up giving users a misleading action item.

I'd bail out of the membership check when the lookup fails or when we don't have a login, and instead add a "Could not verify admin status" warning so we don't surface a false "not a member" message.

```python
org_login = org_identifier
if org_identifier.isdigit():
    org_lookup_response = await client.get(...)
    if org_lookup_response.status_code == 200:
        ...
    else:
        logger.warning(...)
        continue  # bail here so we don’t hit /orgs/<id>/memberships
```

preloop-agent:
`validate_token_permissions` falls back to using the raw `org_identifier` when the `/organizations/{id}` lookup fails and then proceeds to call `/orgs/{org_login}/memberships/{username}`. In our data the identifier is the numeric GitHub org ID, so if the lookup returns 403 (which is very common when the PAT does not have `read:org`) we continue with the numeric ID and the membership call always returns 404. That path now produces the "You are not a member… webhook registration will fail" warning even when the user is actually an admin. We should skip the membership check when we cannot resolve the slug (or downgrade to a "could not verify" warning) rather than emitting a false negative.

preloop-agent:
**Issue:** The backend now returns `warnings` when registering a tracker, but on the happy path we immediately ignore the response payload. In `handleSave` we just fire `tracker-added`/`tracker-updated` after `addTracker`/`updateTracker` resolves, so the user never sees that the token is missing scopes or org admin privileges. That defeats most of the value of the new validation logic.

Can we surface those warnings (toast, alert in the modal, etc.) so the person who just added the tracker knows what to fix? Otherwise they'll only discover the problem later when webhooks fail.

preloop-agent:
Potential bug at frontend/src/components/add-tracker-modal.ts:558

We set `warningMessages` so the modal can stay open and display the warnings, but by the time we reach this block we have already dispatched `tracker-added` / `tracker-updated`. Both `trackers-view` and `flow-view` respond by flipping `isAddingTracker` / `editingTracker` to false, which drops the modal from the DOM immediately. The result is that the warning banner never becomes visible for the user. Could you defer emitting these events until after we decide whether to close the modal (or otherwise make sure the parent doesn’t tear the modal down while warnings need to be shown)?

preloop-agent:
Hi folks—thanks for the updates! I spotted one blocker while reviewing the diff:

• **docker-compose.yml:102** – The new `console` service is built from `./frontend` but still tags the image as `ghcr.io/preloop/preloop:latest`. That is the exact tag the API/worker/scheduler services use for the backend image. As soon as someone runs `docker compose build console`, the frontend build overwrites the backend image tag, so subsequent `docker compose up` tries to start the API containers with the static frontend image and they crash. The fix should be to give the console build its own tag (e.g. `ghcr.io/preloop/console:latest`, which also matches the new CI metadata entries) so the backend and frontend images stay distinct."}

PR #64, opened 2026-07-17
CLI: make agents discover/onboard resilient to per-agent failures

d-mo:
## Problem
A failure onboarding one agent aborted the entire `preloop agents discover` run: remaining agents were never onboarded, and the runtime error printed the full usage/flags dump. Observed in the field on WSL: Codex CLI's config lived in the Linux home but the binary was installed Windows-side (invisible to WSL's PATH) — the launcher error killed the run before Claude Code got its turn. Separately, Claude Code installs that only have `~/.claude.json` (fresh/lightly-used) were never discovered at all.

## Changes
- Per-agent isolation: discover/onboard continues past per-agent errors and prints an end-of-run summary table (`onboarded / partial / failed` + reason). Exit code: 0 if at least one agent onboarded (partial counts), non-zero only if all failed — documented in `--help`.
- Missing agent executable no longer fails that agent: MCP + model routing complete, managed-launcher integration is skipped with a warning; agent is reported as `partial`.
- WSL detection + hint: "agents installed on Windows are not on the WSL PATH — install the agent inside WSL or add its Windows install dir to PATH."
- Runtime errors no longer print the cobra usage dump (usage still prints for genuine flag/argument errors).
- Claude Code discovery extended: `~/.claude.json` and the `claude` binary on PATH are now discovery signals (previously only `~/.claude/settings.json`).

## Testing
`go vet ./...` clean; `go test ./...` green. New table-driven tests: continue-on-error matrix (one fails / all fail / launcher-missing→partial), exit-code convention, WSL detection and hint, Claude Code marker-file matrix, usage-suppression vs genuine flag error (end-to-end through rootCmd.Execute).

## Known follow-up (not in this PR)
`onboard --all`'s control-onboarding path can still abort mid-loop on a non-launcher error — narrower cousin of the same class, left out to keep the diff right-sized.

> [!NOTE]
> **Critical Risk**
> Follow-up commit `f69709e5` introduced a compile error (`agents.go:442: no new variables on left side of :=`) when applying the column-header fix. Must be fixed before the next build.

> **Overview**
> Makes CLI agent discovery/onboarding resilient to per-agent failures: continues past individual agent errors, prints a summary table, degrades gracefully when agent binaries are missing (especially on WSL), suppresses usage dumps for runtime errors, and extends Claude Code detection.

> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit f69709e5. Updates automatically on new commits.</sup>

### DIFF
--- a/cli/internal/cmd/agents.go
+++ b/cli/internal/cmd/agents.go
@@ -438,6 +438,8 @@ func printAgentOnboardingSummary(w io.Writer, outcomes []agentOnboardingOutcome)
 	}
 	fmt.Fprintln(w, "\nOnboarding summary:") //nolint:errcheck
 	tw := tabwriter.NewWriter(w, 0, 4, 2, ' ', 0)
+	fmt.Fprintf(tw, "  Agent\tStatus\tReason\n") //nolint:errcheck
+	tw := tabwriter.NewWriter(w, 0, 4, 2, ' ', 0)
 	missingExecutable := false
 	for _, outcome := range outcomes {
 		reason := outcome.Reason
--- a/cli/internal/cmd/agents.go
+++ b/cli/internal/cmd/agents.go
@@ -6,6 +6,7 @@ import (
 	"crypto/sha1"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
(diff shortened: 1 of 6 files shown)


d-mo:
Rebased onto main across the native-approvals merge (5d225a8d). The only textual conflict was in root.go's PersistentPreRun, resolved by keeping both main's command-usage telemetry increment and this branch's silenceUsageForRuntimeErrors call; main's approval-hook/credential changes slot cleanly into the resilient enrollment path (hook install errors are now classified per-agent instead of aborting the batch). go vet and the full cli test suite pass, including the new resilience tests and main's permission-hook/policy tests.

PR #65, opened 2026-07-17
Installer: survive no-tty/dash, activate HTTPS on cert reuse, reopen registration on fresh DB

d-mo:
## Problems (found by automated end-to-end lifecycle testing of the release install path)
1. `install-oss.sh` dies under dash when no tty is present: the `has_tty` probe's `/dev/tty` redirection is fatal for a special builtin under POSIX sh — any non-interactive invocation (`curl | sh` in CI, provisioning, containers) crashed instead of proceeding unattended.
2. Re-install with an existing certificate left the proxy HTTP-only: `issue_certificate` early-returns when a valid cert exists, so `tls/active.conf` stayed in its HTTP bootstrap state.
3. Re-install over a previous instance's config inherited `REGISTRATION_ENABLED=false` onto a fresh empty database — an instance nobody can register on.

## Changes
1. `has_tty()` contains the `/dev/tty` probe in a subshell — safe under dash and bash, with and without a tty; the installer now runs unattended end-to-end.
2. HTTPS activation extracted into `activate_https()`, called from both the fresh-issuance path and the cert-already-present path.
3. New `reset_registration_if_fresh_db()`: `REGISTRATION_ENABLED=false` now asserts a property of the database ("its first user exists"). After compose-up, a verified user count of exactly 0 reopens registration for the first signup; the existing lockdown flow re-closes it. Fail-closed: any query failure leaves the flag untouched, so a flaky upgrade can never reopen a locked-down instance that has users.

## Testing
Shellcheck clean; `dash -n`/`bash -n`/`sh -n` clean. Verified live on a Debian 12 host over `ssh` + `sh` (dash, no tty): full teardown/reinstall ran unattended to completion; HTTPS active with the reused certificate (same serial before/after — no new issuance); locked `.env` + empty DB produced an instance with registration open; features endpoint agrees at every step.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

> [!NOTE]
> **Low Risk**
> Three well-scoped installer fixes (POSIX tty compatibility, HTTPS cert-reuse activation, fresh-DB registration reset) with fail-closed safety design. Shell-checked, syntax-validated, and live-tested.
>
> **Overview**
> Fixes `install-oss.sh` to survive dash/no-tty environments, activate HTTPS when reusing existing certificates, and detect fresh databases to re-enable registration — ensuring CI/provisioning installs work unattended and re-installs produce working instances.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit fix/installer-unattended-tls-registration. Updates automatically on new commits.</sup>

### DIFF
--- a/scripts/install-oss.sh
+++ b/scripts/install-oss.sh
@@ -317,7 +317,11 @@ reset_registration_if_fresh_db() {
     cd "$INSTALL_DIR"
     # shellcheck disable=SC2086
     docker compose $COMPOSE_ARGS up -d api > /dev/null 2>&1
-  ) || true
+  ) || {
+    echo "  WARNING: could not restart the API container;" >&2
+    echo "  REGISTRATION_ENABLED=true may not be applied. Run manually:" >&2
+    echo "  cd ${INSTALL_DIR} && docker compose ${COMPOSE_ARGS} up -d api" >&2
+  }
 }
 
 # Email is how approvals, invitations and password resets reach people. Ask
--- a/scripts/install-oss.sh
+++ b/scripts/install-oss.sh
@@ -128,8 +128,14 @@ env_value() {
 # True only when a real terminal is attached. `curl ... | sh` leaves stdin as
 # the pipe, so prompts must go through /dev/tty — but /dev/tty merely EXISTING
 # is not enough (in CI / cloud-init it opens and then fails), so probe it.
+# The probe must run in a subshell: `:` is a POSIX special builtin, and a
+# failed redirection on a special builtin is FATAL under sh implementations
+# like dash — a bare `{ : < /dev/tty; }` kills the whole script the moment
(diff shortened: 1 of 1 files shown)


d-mo:
Applied the review suggestion in 49073a2a: the `|| true` on the `docker compose up -d api` restart inside `reset_registration_if_fresh_db` now emits a stderr warning instead of silently swallowing failures. The install still never aborts, but the operator is told that REGISTRATION_ENABLED=true may not be applied and is given the exact manual command including the compose file args (`cd ${INSTALL_DIR} && docker compose ${COMPOSE_ARGS} up -d api`). Re-verified: shellcheck (koalaman/shellcheck:stable) clean, `bash -n` / `sh -n` / `dash -n` clean, and the fallback path exercised under dash (warning on stderr, exit 0).

PR #78, opened 2026-07-18
Launch-week hardening: push, optimization, Windows, and plugin fixes

d-mo:
Groups the launch-week fixes. Each section is independent; the push and Hermes ones are the load-bearing changes.

## Push notifications were fully broken on prod and staging

Three defects composed. The Android app registered a literal `fcm_unavailable_<millis>` placeholder when the Firebase token fetch failed — non-blank, so it passed every backend check before FCM rejected it as an invalid registration token. It could never be pruned, because error classification matched on substrings that miss `INVALID_ARGUMENT`. And it could never be replaced, because the token-refresh endpoint 404'd inside a log-only catch. QR pairing used a different, correct path, so pairing looked healthy while delivery was dead.

Backend now classifies FCM errors by type, prunes only client-side faults (a credential outage must not wipe every user's token), logs the Firebase `project_id`, and rejects placeholder-shaped tokens at registration. Adds an admin-only test-send that exercises the real provider path and surfaces the verbatim error — the synthetic approval is persisted nowhere, so it cannot leak into lists, feeds, or metrics.

## The Hermes plugin gated nothing

`pre_tool_call` was `async def`, but Hermes invokes hooks synchronously and discards non-dict results, so every tool call proceeded ungated. Verified against real Hermes source, and reproduced:

```
OLD (async def registered)   -> action=None     *** TOOL RAN UNGATED ***
NEW (sync bridge registered) -> action='block'  BLOCKED
```

Registers a sync bridge over a dedicated loop, correct from all three contexts Hermes calls hooks from. Two further silent-allow paths also now block: unreadable config, and non-mapping response bodies.

Note the behavior change at the install cliff — a user who installs but has not run `login` now has tool calls blocked with an actionable message rather than running ungoverned. That matches the documented fail-closed default, but it is stricter than before.

## Session optimization

Savings could exceed analyzed scope (131% observed): schema tokens were already resend-aware, then multiplied by `resend_count` again, making `scope-tools` savings quadratic. Adds a deduped profile-level roll-up and clamps totals to analyzed scope with a logged warning.

Default model selection now excludes principal-bound OAuth credentials, which cannot serve server-side generation — previously a Claude-Code-only user's first Optimize click failed server-side.

Adds a bundled example session so a first-time Optimize tab shows real suggestions instead of an empty state. It runs the production analyzers over an in-memory transcript with zero DB writes, so it cannot contaminate cost or savings aggregates.

## Windows

The OSS stack already ran under Docker Desktop with WSL2 and Windows CLI binaries already shipped on every release — neither was documented. Adds that, fixes Claude Desktop discovery via `UserConfigDir` (which also picks up `~/Library/Application Support` on macOS), fixes MSYS path mangling in the installer, and adds a non-gating `windows-latest` CI job.

The CI job will be red. That is deliberate and informative: ~90-100 test functions fail because they use `t.SetEnv("HOME", ...)` while Windows reads `%USERPROFILE%`, so they read and write the runner's real `~/.preloop`. Triage is in the job's comment block.

## Release and packaging

`release.py` now restamps `openapi.yaml`, which went stale on every version bump and failed the lint job. PyPI metadata was completely empty (no keywords, license, URLs, or classifiers) and is now populated. OpenClaw plugin goes to 0.2.0, with ClawHub publishing automated alongside npm and guarded by a digest comparison — the two registries had shipped different artifacts under 0.1.1.

> [!NOTE]
> **Low Risk**
> All previously flagged issues have been addressed. Only minor code-quality observations remain. Critical push notification triple-defect, Hermes sync bridge, and session optimization clamping are well-tested.
>
> **Overview**
> Launch-week hardening across 5 areas. Push notification triple-defect fix (token validation + FCM error classification + refresh endpoint), Hermes plugin sync bridging, session optimization clamping + OAuth model exclusion + bundled example session, Windows docs + CI, and release/packaging fixes. 30 files, 4 new test files with 900+ lines. 8 of 8 previously flagged issues resolved.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 19e455bc. Updates automatically on new commits.</sup>


d-mo:
Addressed the code-quality findings in `f9d128cf`, `0520ae79`, and `5167595c`.

**Implicit string concatenation (`generate_example_session.py`) — valid, fixed.** Both lines were split for width, which does read as a missing comma. Wrapped in explicit parentheses. Verified the generated fixture is byte-identical by hash, so the change is purely cosmetic.

**Implicit operand conversion (`session-replay-panel.ts`) — false positive, but it pointed at a real design smell, so it is fixed properly.** `.prop=` is a Lit *property* binding rather than an attribute, and both targets are declared `| null`, so no string conversion occurs. Naming typed constants did not satisfy the analyzer either, because it reads the Lit template as string interpolation regardless.

Rather than suppress it, I removed the reason the nulls were there. They existed only to stop Lit reusing DOM across renders and leaking a real session's values into the example view. Keying the panel gives it a distinct identity directly, so the properties stay at their null defaults, the bindings disappear, and the warning goes with them. Better code than what I had — arrived at by taking the flag seriously rather than dismissing it twice.

### Windows CI, for context

`CLI Tests (Windows)` is now green for the first time. 95 test call sites across 16 files were setting `HOME`, but production calls `os.UserHomeDir()`, which reads `%USERPROFILE%` on Windows — so the redirect did nothing there and the tests were reading and writing the runner's real `~/.preloop`. The new helper also sets `APPDATA`/`LOCALAPPDATA`, since `os.UserConfigDir` reads those directly and does not derive from home.

Fifteen tests skip on Windows with explicit reasons; five of those are labelled a product gap rather than a test artifact, because `preloop agents onboard` genuinely cannot install a managed launcher on Windows today. A canary test fails if someone implements support, so those skips cannot rot silently.

d-mo:
All five remaining review items addressed in `19e455bc`.

**[LOW — Maintainability] Escalation/main push duplication** — extracted the shared piece: `_send_android_push_transport` is now the single place that knows both transports' call signatures, used by both loops. That is exactly the axis that drifted (`device_token=` against a `token` parameter). Result *policy* deliberately stays per-caller — the main path prunes dead tokens, escalation only counts, since the main path prunes the same token on its next regular send. The iOS branches are intentionally not merged: their payloads genuinely differ, and the divergence class that produced the bug was transport signatures. Two corners tidied along the way: with no transport configured, the main path records a clean failure instead of calling an unconfigured proxy, and escalation counts it as failed instead of skipping silently.

**[LOW — Thread Safety] `_initialize_fcm` globals** — double-checked lock around first-time init. The race was benign as noted, but benign races rot into real ones when someone adds a second global.

**[LOW — Resource] `_test_push_state` growth** — swept opportunistically once per window on the next call; no timer thread for an admin-only surface. Two new tests cover eviction and that the caller's own fresh entry survives a sweep.

**[LOW — Observability] Fire-and-forget pruning** — the executor future now has a done-callback, so a crash before the function's own try block (session factory, executor shutdown) is logged instead of vanishing. Also commits per token now, so one bad row cannot roll back the batch and keep every other dead token alive.

**[MEDIUM — Design] QR token in the deep-link query string** — documented as a deliberate decision rather than changed, for two reasons the review did not have visibility into: the token is **single-use** (`validate_and_consume` in `qr_service.py:86` burns it on first registration) and expires after 15 minutes, so a logged copy is dead the moment pairing completes; and the shipped iOS/Android apps parse exactly this URL shape, so changing the format breaks pairing for every installed app version. The threat model is now written at the call site so the next reader does not have to rediscover it.

160 tests across the affected suites pass; ruff clean.

PR #88, opened 2026-07-19
Launch hardening: bootstrap registration, async optimize jobs, model authorization, funnel telemetry

d-mo:
Four hardening lanes for the 0.12.x launch cut, plus a small docs sweep. Combined-tree verification: backend 4087 passed / 6 skipped (integration tests excluded — they need a live stack), Go CLI all packages pass, frontend 542 passed, EE plugin halves tested separately (111 passed).

## Bootstrap registration token
Closes a first-signup takeover race on fresh public instances. With zero users and `PRELOOP_BOOTSTRAP_TOKEN` set, `/register` requires the token (constant-time compare; pg advisory lock serializes the first signup; account+user+role commit in one transaction). The installer generates the token, persists it to the instance `.env`, and prints a `/register#bootstrap=<token>` setup link to the terminal only (never install.log). The register view reads the fragment, strips it from history, and shows 'setup link required' / 'already claimed' notices.

## Async optimization jobs
Replaces the ~90s synchronous wait on session optimization with a background job: new `OptimizationJob` model + migration (partial unique index makes double-submit idempotent), bounded `ThreadPoolExecutor(3)` off the event loop with per-job DB sessions and heartbeats, a sweeper for stalled jobs, `POST …/optimizations/jobs` → 202 + polling. Console shows analyzing / failed+retry / no-waste states with full a11y. Legacy sync endpoint unchanged.

## Per-principal model authorization
One `compute_authorized_model_ids()` consumed by model listing, requested-model resolution (exact + suffix — closes the hidden-alias hole), and default selection. BYOK models visible to all; principal-bound OAuth models only to their bound managed agent; fail closed on unbound credentials. 400s enumerate usable model ids (`code=model_not_authorized`) for both OpenAI and Anthropic envelopes.

## Observable funnel telemetry
`install_completed` rides the existing daily version check-in (once, persisted, suppressed when telemetry disabled or hosted). `cli_first_run` fixes a check.go bug where first_run read false on the very first run. `first_session_seen` is a new OSS session-events hook registry — inert no-op unless a plugin registers a handler; the event never leaves the instance. Payload contract documented in SECURITY.md.

## Docs
Standardize `preloop login` examples on the console origin `:3000` (console nginx proxies `/api`).

> [!NOTE]
> **Low Risk**
> Four independent hardening lanes, each well-tested with sound security design and proper isolation.
>
> **Overview**
> Bootstrap registration token prevents first-signup takeover races with constant-time compare and advisory locks. Async optimization jobs replace the ~90s synchronous wait with a bounded thread pool, heartbeat monitoring, and recovery sweeper. Per-principal model authorization closes the hidden-alias hole with a single authorization chokepoint. Telemetry hooks are opt-out and EE-isolated with full contract documentation in SECURITY.md.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit launch-hardening-0.12.7. Updates automatically on new commits.</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Bootstrap setup token for the first registration.** On a fresh (zero-user)
+  instance with `PRELOOP_BOOTSTRAP_TOKEN` configured, `/register` requires the
+  token: the installer generates one, persists it to the instance `.env`, and
+  prints a `/register#bootstrap=<token>` setup link to the terminal only. This
+  closes the race where a stranger could claim a freshly installed public
+  instance before its operator. First signup is serialized with a database
+  advisory lock, and account+user+role now commit in a single transaction.
+- **Async session-optimization jobs.** `POST /account/runtime-sessions/{id}/optimizations/jobs`
+  runs the analysis in a bounded background worker and returns `202` with a
+  pollable job; the console Optimize tab shows analyzing / failed+retry /
+  no-waste states instead of a spinnerless multi-minute wait. The synchronous
+  endpoint is unchanged.
+- **Activation telemetry markers.** Self-hosted instances report a one-time
+  `install_completed` marker on the existing daily version check-in, and the
+  CLI reports a one-time `cli_first_run` marker — both suppressed by
+  `PRELOOP_DISABLE_TELEMETRY`. A new in-instance `first_session_seen` hook
+  registry lets plugins observe first agent activity; the event never leaves
+  the instance. The full contract is documented in SECURITY.md.
+
+### Fixed
+
+- **Per-principal model authorization at the gateway.** Model listing,
+  requested-model resolution (exact and suffix), and default-model selection
+  now all consume one authorized-model computation: principal-bound
+  subscription-OAuth models are visible only to their bound managed agent, and
+  credentials with no bound models fail closed instead of seeing everything.
+  Rejections return `model_not_authorized` with the usable model ids.
+- CLI first-run telemetry read `first_run=false` on the very first run.
+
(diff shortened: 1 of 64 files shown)


d-mo:
Review addressed in df437d0d:

- **Lint failure**: openapi.yaml was stale for the new endpoints. Root cause was in `scripts/generate_openapi.py` itself — it put the repo root on `sys.path`, but the package root is `backend/`, so the import silently fell through to site-packages and generated from whatever checkout the venv's editable install pointed at. The script now pins `backend/` ahead of site-packages, and the schema is regenerated from branch code.
- **CHANGELOG**: entries added for all four lanes under Unreleased.
- **gemini_gateway.py discarded return**: comment added — the call exists to raise the shared `model_not_authorized` 400.
- **bootstrap.py `_users_exist_state` dict-cache**: keeping as is. The in-code comment explains it: a module-level `global` flag trips a CodeQL unused-global false positive; the mutable-mapping form avoids that while keeping the sticky-True semantics and the test reset hook.

d-mo:
Three more commits pushed:

- **f178743f** — same sys.path pin applied to the remaining operator scripts (`create_first_user.py`, `manage_superusers.py`, `validate_subscriptions.py`) so none of them can silently run another checkout's code.
- **73f561b7** — **subscription-login custody fix.** Claude Code / Codex OAuth refresh tokens are single-use and rotate on every gateway-side refresh, so the copy left in the agent's local credential store dies while onboarded, and offboard cleanup could delete the only live copy — costing the operator their login. New `POST /api/v1/ai-models/{id}/credentials/export` (subscription-OAuth types only, refreshed if stale, owner-authenticated, 404 across accounts, never for API keys) + offboard now writes the live bundle back to `~/.claude/.credentials.json` / Codex `auth.json` (unrelated fields preserved) before cleanup. 7 new backend tests, 5 new CLI tests.
- **8853a113** — CLI 401s now hint `preloop login` instead of a bare API error (`--token` sessions expire silently).

PR #89, opened 2026-07-19
Route unknown-provider models through litellm's OpenAI-compatible adapter

d-mo:
Fixes the Hermes onboarding live-validation failure reported by an external tester today (0.12.7):

```
litellm.BadRequestError: LLM Provider NOT provided. You passed model=kimi-for-coding/k3
```

Hermes treats `model.provider: custom` as any OpenAI-compatible endpoint, so imported provider names can be arbitrary (`kimi-for-coding`). The gateway forwarded that name as a litellm prefix, and litellm rejects unknown providers even when `api_base` is set.

Resolution order in `_to_litellm_model` is now:
1. Local prefix map (unchanged).
2. Native litellm providers (mistral, groq, …) pass through — unchanged behavior, now explicit.
3. Unknown provider **with** an endpoint → `openai/<model>` + the model's `api_base` (litellm's generic OpenAI-compatible route).
4. Unknown provider without an endpoint → keep the name so the error stays attributable.

4 new tests. Note: `policy_generation.py`, `runtime_session_explorer.py`, and `approval_summary.py` carry near-copies of this builder that would hit the same bug when an account's default model is a custom-provider one — proposing a shared-helper unification as post-launch cleanup rather than bundling it here.

> [!NOTE]
> **Low Risk**
> All previously raised issues resolved. Shared `litellm_routing.py` consolidates model-string building across 5 call sites with comprehensive test coverage.
>
> **Overview**
> Expands the original `kimi-for-coding/k3` routing fix into a shared `litellm_routing.py` module used by gateway, policy generation, approval summaries, session explorer, and account.py. Four divergent builder copies are now consolidated. Callers of custom-provider models (Hermes-imported endpoints) route through litellm's generic `openai/` adapter. All findings from the previous review have been addressed.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit d16902d9. Updates automatically on new commits.</sup>


d-mo:
Pushed d16902d9 — "fix them all" follow-through:

**Backend — one shared litellm routing module.** `preloop.services.litellm_routing.to_litellm_model` replaces the four divergent builder copies (gateway, policy generation, approval summaries, session explorer), so internal features no longer break when an account's default model is a custom OpenAI-compatible provider. Pass-through of prefixed identifiers now requires the head to actually be routable (`azure/gpt-5.4` passes; `meta-llama/llama-3.1-70b` still gets its provider prefix). Full backend suite: 4101 passed.

**CLI — npm-tarball fallback for `install-plugin`.** OpenClaw's ClawHub download path fails client-side (ENOENT on the zip) even now that 0.2.0 is live on ClawHub with matching shasums — verified on a live VM. When the marketplace install errors, the CLI now fetches the npm tarball and installs it as a local file (OpenClaw's local-tarball path works), mirroring the Hermes pip fallback. Verified manually end-to-end on the VM; 4 new Go tests.

d-mo:
Review addressed in f8b214a0:

- **Critical — `account.py:1207` broken lazy import**: confirmed and fixed; `extract_agent_name` now uses the shared `to_litellm_model` (which also fixes its behavior for custom-provider default models). Added an endpoint-level regression test that exercises the lazy-import path — module-import checks alone can't catch these.
- **Scope creep**: the OpenClaw npm-tarball fallback is split out to its own PR.
- **CHANGELOG**: Unreleased entry added for the custom-provider routing fix.

Routing test suite: 7 passed. CLI reverted to main state on this branch and still builds.

Issue #123, opened 2026-07-30
Feature: ingest Cursor bundled-model usage into Cost analytics


ajlennon:
## Summary

Cursor onboard gives MCP firewall / approvals, but **bundled Cursor models never hit the Preloop model gateway**, so their spend does not appear in Cost analytics or budgets.

Docs already call this out ([Cursor, Claude Desktop & Others](https://docs.preloop.ai/clients/other-mcp-clients/)): model routing for Cursor is manual OpenAI base-URL only, and that path does not cover Cursor’s own / bundled models used by Agent/Composer day-to-day.

## Problem

For users who mostly run Cursor bundled models:

1. `preloop agents onboard cursor` → MCP proxy only (validated).
2. Manual `https://…/openai/v1` override does not capture bundled-model traffic.
3. Cost / Runtime Sessions stay empty for that spend, while tool governance still works.

That leaves a hole exactly where a lot of real agent spend lands.

## Ask

Support **Cost visibility for Cursor bundled-model usage** without requiring gateway interception of Cursor’s proprietary model path. Concrete options (any one, or a mix):

1. **Official ingest API** — accept normalized usage events (timestamp, model, tokens or charged cents, optional session/agent id) into `ApiUsage` (or a sibling ledger) attributed to a managed Cursor agent / runtime principal.
2. **CSV import** — import Cursor dashboard Usage export (per-request rows Cursor already exposes in the UI) into Cost with attribution.
3. **Documented connector** — first-party or community adapter that pulls Cursor usage (prefer official Admin/Teams APIs when available; otherwise a documented, opt-in dashboard export path) and writes into Preloop Cost.

Nice-to-haves:

- Attribute rows to the existing managed Cursor agent from `preloop agents onboard`.
- Keep gateway budgets separate from “imported/observed” spend (or clearly label `usage_source` / `cost_source`).
- Avoid requiring reverse-engineered cookie auth as the only path.

## Context / evidence

- Local Cursor enrollment: MCP onboard OK; validate reports model traffic still direct.
- Manual OpenAI base URL is chat-oriented and insufficient for bundled Agent models.
- Cursor surfaces usage in-product (Settings → Usage Summary; dashboard Usage / CSV). Programmatic access for individuals is weak/unofficial; Teams/Enterprise Admin APIs exist for orgs.

## Workaround today

Track bundled spend in Cursor’s own dashboard/CSV; keep Preloop for MCP governance/audit only.

Happy to test a prototype ingest against Cloud if useful.

[...]

PR #167, opened 2026-08-04
feat(cli): add `preloop usage import` for externally observed spend (#123)

d-mo:
## What

Adds `preloop usage import <file>`, the CLI path to the usage ingest
endpoints that landed in #137. Refs #123.

```
preloop usage import cursor-usage.csv
preloop usage import events.json --agent-id 7f1c...c3
preloop usage import export.csv --column-map '{"cost":"Cost to You"}'
```

## Why

#137 shipped `POST /api/v1/usage/import` (normalized JSON events) and
`POST /api/v1/usage/import/csv` (Cursor dashboard Usage export) so that
Cursor's bundled Composer/Auto spend, which never traverses the model
gateway, can appear in Cost analytics. Until now the only way to reach
either endpoint was hand-rolled curl, which does not match the real
workflow: export a CSV from the Cursor dashboard, then import it.

## How

The file extension picks the endpoint. `.csv` is posted as multipart to
the CSV endpoint; `.json` is parsed locally and posted as normalized
events. An unknown extension is an error rather than a content sniff,
because sending a mislabeled file to the wrong parser fails in a much
more confusing way than a rename does.

Flags mirror the API surface: `--agent-id`, `--source` (default
`cursor`), and `--column-map` for CSV shapes the built-in header matcher
does not recognize. `--column-map` is validated locally (JSON syntax, and
rejected outright for JSON input) so an obvious mistake does not turn
into a server-side 422.

Two behaviors worth review attention:

- JSON files are split into batches of 5000 events, matching
  `MAX_EVENTS_PER_REQUEST` in `backend/preloop/schemas/usage_import.py`,
  and the per-batch counters are summed. Without this a large export just
  fails validation. The import is idempotent, so a partially applied
  split can be re-run safely.
- Both the bare array and the `{"events": [...]}` envelope are accepted,
  since both are shapes people plausibly have on disk.

The summary prints imported and duplicate counts plus the attributed
agent, and surfaces the CSV endpoint's `skipped_row_reasons`. API errors
are wrapped rather than replaced: the 422 an operator hits most often
names the `preloop agents onboard cursor` prerequisite, and swallowing
that body would hide the fix.

## Testing

`cli/internal/cmd/usage_test.go` covers extension routing, both JSON
input shapes and their rejections, request shape for JSON and multipart
(including that unset optional fields are omitted so the server resolves
the default agent), batching of oversized JSON files with summed
counters, error-body pass-through of the onboard-prerequisite 422,
summary rendering with and without skipped rows, flag validation, and
command registration.

Full CLI suite green locally (`go test ./...` in `cli/`).

## Not in this PR

- CHANGELOG: the current `[Unreleased]` section is the 0.14.0 content
  owned by #164, which renames that heading. Adding a second heading here
  would conflict during the cut. This entry should go under a fresh
  `[Unreleased]` after 0.14.0 is tagged.
- Docs coverage lives in the separate `preloop-docs` repo and is tracked
  in a companion MR there.

## Note on merge

Do not merge before the 0.14.0 cut. This is post-0.14.0 work.

> [!NOTE]
> **Low Risk**
> Well-tested CLI addition with no security concerns, proper input validation (including column-map shape and field-name validation), and comprehensive error handling.
>
> **Overview**
> Adds `preloop usage import <file>` for CSV and JSON usage ingest. File extension routing, local flag validation with field-name checking, JSON batching for large files, and full API error passthrough. 557 lines of tests covering all paths.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit c8d377b5. Updates automatically on new commits.</sup>

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -119,6 +119,17 @@ preloop tools exec <tool-name> --args-file ./input.json
 
 `preloop tools` talks directly to the MCP endpoint, so the visible and executable tools are automatically filtered by the current token's policy. Agent tokens only see the tools they are allowed to use.
 
+### Usage Import
+
+```bash
+preloop usage import cursor-usage.csv                # Cursor dashboard Usage export
+preloop usage import events.json                     # Normalized usage events
+preloop usage import cursor-usage.csv --agent-id <id>
+preloop usage import export.csv --column-map '{"cost":"Cost to You"}'
+```
+
+Imports spend the model gateway never saw, such as Cursor's bundled models. Records are labeled as imported, so they are reported separately from gateway-metered spend and never count against gateway budgets. Re-importing the same file is safe: duplicates are detected and reported as skipped. Without `--agent-id`, the account's onboarded Cursor agent is used, so run `preloop agents onboard cursor` first or pass the id explicitly.
+
 ### Approvals
 
 ```bash
--- a/cli/internal/cmd/root.go
+++ b/cli/internal/cmd/root.go
@@ -110,5 +110,6 @@ func init() {
 	rootCmd.AddCommand(toolsCmd)
 	rootCmd.AddCommand(approvalsCmd)
 	rootCmd.AddCommand(agentsCmd)
+	rootCmd.AddCommand(usageCmd)
 	rootCmd.AddCommand(versionCmd)
 }
--- a/cli/internal/cmd/usage.go
+++ b/cli/internal/cmd/usage.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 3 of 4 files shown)


d-mo:
Addressed the review in c8d377b5.

**Medium: `--column-map` shape validation**: fixed, and the finding was right about why it mattered. `json.Valid` accepts `null`, `42` and `[1,2,3]`, so the local check was waving through exactly the inputs it existed to catch, and the operator got a server-side 422 anyway.

It now parses as `map[string]string`, which rejects scalars, arrays, and non-string values with a message showing the expected form. I went one step further than the suggestion and also validate the field *names* against the server's `_CURSOR_LOGICAL_FIELDS` list, since the server rejects an unknown field outright rather than ignoring it. A typo like `{"price":"Cost"}` now gets told which names are valid, which is the part that actually unblocks someone. An empty map is rejected too: it is a no-op the caller did not intend. A test asserts every field the CLI accepts is in that list, so it cannot drift from the server's silently.

**Low: nil safety in `writeUsageImportSummary`**: added the guard. The caller does guarantee non-nil today, but that is one refactor from being untrue and a panic is a poor way to discover it.

**Low: discarded `GetString` errors**: left as is. `Flags().GetString` only errors when the flag is not defined, which is a programming error caught by `TestUsageImportCommandIsRegistered`, not a runtime condition. Handling it would add noise and diverge from every other command in `cli/internal/cmd`.

Full CLI suite green locally.

PR #180, opened 2026-08-05
fix: resolve aux model credentials via secret service, add system-default fallback

d-mo:
## Problem

Seven auxiliary generation sites read the raw `ai_models.api_key` column directly
instead of resolving credentials through the secret service. Any account whose
model is configured with `credentials_secret_id` (vault-backed) has an empty
`api_key` column, so these paths sent requests with no key and got silent 401s:

- `services/approval_summary.py` (approval summaries)
- `services/runtime_session_explorer.py` (session interaction titles)
- `services/policy_generation.py`
- `api/endpoints/account.py` (agent name extraction)
- `api/endpoints/issue_compliance.py`
- `api/endpoints/issue_duplicates.py`
- `api/endpoints/issue_dependencies.py`

The main gateway already did this correctly via
`SecretService.resolve_ai_model_credentials`. Seven hand-rolled copies of the
credential block is how the drift happened, so this adds one shared helper rather
than seven copies of the fix.

## Part 1: shared credential resolution

New `services/model_credentials.py` exposes `resolve_model_call_credentials(model, db=...)`,
returning a litellm/openai kwargs fragment. It wraps the same resolver the gateway
uses, so vault-backed secrets, legacy plaintext keys, OAuth, and ambient credentials
all behave identically to the gateway path. An `api_key` is only injected when the
credential type is `api_key`, matching `model_runtime_resolver`.

Two behavior details worth review:

- `api_base` is set outside the try block. Routing does not depend on credentials,
  and folding it inside meant a secret backend failure silently dropped the custom
  endpoint and sent the request to the default provider. Caught by
  `test_extract_agent_name_uses_shared_routing`.
- The helper never raises. These are all best-effort auxiliary paths, so on
  resolution failure it logs and omits the key, degrading as before.

The `os.getenv("OPENAI_API_KEY")` last-resort in the compliance and duplicates
endpoints is preserved, and only consulted after the resolver returns nothing.
Existing `HTTPException` contracts are unchanged.

## Part 2: call-time fallback to the system default

Scope is deliberately narrow: auxiliary, non-interactive, UX-affecting generations,
specifically approval summaries and session interaction titles. When the account's
model fails, these retry once with the system-wide default model
(`account_id IS NULL`), which is read as data. No model names are hardcoded.

Guardrails:

- Per-account, per-UTC-day cap via `PRELOOP_AUX_FALLBACK_DAILY_CAP` (default 50).
  The cap is read at call time rather than frozen at import, so it is configurable
  without a rebuild, and a cap of 0 disables fallback entirely.
- The counter slot is claimed under the same lock that checks it, so concurrent
  callers cannot both pass the check and overshoot the cap.
- Deduped per-account daily warning, so a broken model config surfaces as a support
  signal without log spam.
- Per-day pruning keeps the in-process maps bounded.
- Credential resolution runs inside the worker thread, since a secret backend lookup
  can do network I/O and must not block the event loop.
- Each attempt gets its own timeout budget. Previously a hanging primary would
  consume the entire deadline and the fallback would never run.

If the fallback also fails, or no system default exists, or the default is the same
model that just failed, the result degrades exactly as it does today. Nothing new is
raised into callers.

The cap and dedupe state are per-process. In a multi-worker deployment each worker
tracks its own counter, which is acceptable for these paths and noted in the module
docstring.

## Not in scope

The main gateway/completion path is untouched and never falls back. Tests assert
this both by name and at the source level, since a name-only check would pass if
someone imported the module and called it via an attribute.

These auxiliary paths do not meter through the billing hosted-budget today, and this
PR does not add new metering. Fallback generations consume the system default model's
budget. Worth a follow-up if fallback volume grows.

## Testing

`backend/tests/services/test_model_credentials.py` (new, 26 tests) covers credential
resolution per credential type, fallback success and exhaustion, absent and identical
system default, cap enforcement including a zero cap, warning dedupe, day pruning,
off-event-loop resolution, per-attempt timeout, and gateway isolation.
`test_approval_summary.py` gains a test proving a `credentials_secret_id` model
resolves the vault value, which is the actual bug.

Full run of `backend/tests/{services,endpoints,api}` against a clean `ef4b38bb`
baseline worktree:

- baseline: 180 failed, 2680 passed
- this branch: 180 failed, 2707 passed
- new failures: zero, the 180 are pre-existing and unrelated

Three `runtime_session_explorer` tests and one routing test needed updating for the
changed internal signature; the routing one surfaced the `api_base` bug above.

Target release: 0.14.0

> [!NOTE]
> **[Low Risk]**
> Centralized credential resolution and fallback mechanism is well-tested. One minor inconsistency remains where `base_url` is not forwarded in compliance check and duplicate classification functions, but this is a pre-existing pattern and does not block merge.

> **Overview**
> Centralizes AI model credential resolution into a shared `resolve_model_call_credentials` helper and adds a call-time system-default fallback for approval summaries and session titles with per-account daily caps, deduped warnings, and per-attempt timeouts. Gateway path is untouched and verified at source level.

> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit ebcc0024. Updates automatically on new commits.</sup>


d-mo:
All three review findings addressed in `ebcc0024`.

**HIGH: `issue_compliance.py:272` and `issue_duplicates.py:895`.** Both fixed. You were right that these were a different shape from the seven in the original pass: they did not read the raw `api_key` column, they built `openai.OpenAI()` with no credential attempt at all, so they silently used whatever ambient `OPENAI_API_KEY` the process happened to have and failed outright when it had none, regardless of the account's configured model. Both now use `resolve_model_call_credentials` exactly like their already-fixed siblings in the same files, keep the `OPENAI_API_KEY` env last resort, pass the model's `api_base`, and now raise a 500 before dispatch when no key resolves rather than calling out unauthenticated.

**MEDIUM: `issue_dependencies.py:169`.** Converted to the shared helper, consistent with `extend_dependency_scan` in the same file, and `base_url` is now passed. Worth flagging that this was not only a consistency issue: the old direct `resolve_ai_model_api_key()` call resolved the key but dropped `api_endpoint`, so a model with a custom endpoint had its traffic sent to the default provider. The existing regression test now asserts `base_url` is forwarded.

**DOC: `PRELOOP_AUX_FALLBACK_DAILY_CAP`.** Documented in `ARCHITECTURE.md`, in the Runtime Session Identity section next to the sibling `billing_session_*_daily_cap_usd` caps, in a new "Auxiliary Model Credentials & Fallback" bullet that also covers the resolver and the per-process cap semantics.

**Full re-grep, both patterns.** Repeated across the whole backend rather than the single pattern the first pass used:
- Raw `model.api_key` reads outside the resolver: **none**. Remaining `.api_key` hits are the `ApiKey` model/relationship, tracker credentials, gateway auth context, and `secret_service.py` itself (which is the resolver).
- `openai.OpenAI()` / `AsyncOpenAI()` constructions near model usage: **12 total, all resolver-backed**. Nine resolve through `resolve_model_call_credentials`; the other three (`embedding.py:376`, `audio_model.py:72`, `ai_approval_service.py:376`) take an already-resolved key as a parameter from their callers. No bare no-credential construction remains.

**Tests.** 4 new: vault-backed (`credentials_secret_id`) resolution and the no-credential 500 path for each of the two endpoints. Each was verified to fail against the pre-fix code, so they cannot pass vacuously. The dependencies test was updated for the helper and the new `base_url` assertion.

**Verification.** Same worktree, changes stashed vs applied, identical command: 180 failures on both sides with **identical failure sets** (`comm` diff empty, zero new failures), passing 2707 -> 2711. The 180 are pre-existing and unrelated. `ruff check` and `ruff format` clean.

CHANGELOG updated: the entry now says nine sites and calls out the ambient-key behaviour and the dropped-endpoint fix, since both are user-visible.

Not merged, as instructed.

PR #190, opened 2026-08-06
fix: session tracking for all agents, public session semantics, codex flow runtime, aux reasoning knob

d-mo:
fix: session tracking for all agents, public session semantics, codex flow runtime, aux reasoning knob

> [!NOTE]
> **Low Risk**
> No security concerns, well-tested (4,861 tests), already documented in ARCHITECTURE.md and CHANGELOG.md.
>
> **Overview**
> Extends per-agent session splitting to Codex, OpenCode, and OpenClaw; adds an inactivity closer for signal-less agents; fixes Codex tool compatibility in the gateway; adds aux reasoning knob + DB trigger. 21 files changed.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 653b71d. Updates automatically on new commits.</sup>


d-mo:
Thanks — addressed in `767d200e`, with one correction.

**CHANGELOG: valid, done.** Added three `### Fixed` entries under `[Unreleased]` covering the per-agent session fix, the inactivity closer, and the Codex flow runtime.

**Two corrections to the finding, both verified against the source rather than assumed:**

1. **`PRELOOP_AUX_OPENROUTER_REASONING_DISABLED` does not exist.** `grep` across `backend/` returns no such name. The aux OpenRouter knob has **no env var and no opt-out**: `_get_reasoning_disable_defaults` / `get_aux_openai_sdk_extra_kwargs` detect OpenRouter automatically (`provider == "openrouter" or is_openrouter_endpoint(endpoint)`) and always send `extra_body={"reasoning": {"enabled": False}}`. Documenting a configuration flag that isn't there would be worse than documenting nothing, so I didn't. The aux change is also internal-only (auxiliary generations, never the gateway path), which is why it stays out of the changelog.

2. **The idle knob's env var is `RUNTIME_SESSION_IDLE_TIMEOUT_MINUTES`, not `PRELOOP_`-prefixed.** `Settings` sets no `env_prefix`, so the field name maps directly. Verified by actually running it:

   ```
   RUNTIME_SESSION_IDLE_TIMEOUT_MINUTES=99 python -c "from preloop.config import settings; print(settings.runtime_session_idle_timeout_minutes)"
   -> 99
   ```

   (`PRELOOP_AUX_FALLBACK_DAILY_CAP` is prefixed because it's read via an explicit `os.getenv`, not through `Settings` — hence the inconsistency.) The changelog documents the correct, tested name.

`gen_ai.conversation.id` is mentioned in the session-tracking entry as the telemetry vocabulary we follow.

PR #194, opened 2026-08-08
fix: prevent DB pool exhaustion and log-batch data loss

d-mo:
## Problem

Under a burst of concurrent flow work, a gateway pod exhausted its SQLAlchemy connection pool:

```
TimeoutError: QueuePool limit of size N overflow M reached, connection timed out, timeout 30.00
```

The cascade, in order:

1. `websocket_manager._sync_batch_insert_logs` failed and **dropped a batch of NATS logs** (data loss)
2. `Database health check failed` (`health.py:52`)
3. `preloop.api.auth.jwt.get_user_from_token_if_valid` failed
4. Pod restart (`exitCode: 137, reason: Error`, not OOMKilled)

The database itself was nearly idle throughout. This was per-pod pool starvation, not a global connection shortage.

## Root cause

Two distinct problems, both fixed here.

**1. Per-pod pool sizing did not account for two engines per pod.**
`models/db/session.py` passes the same `_database_pool_kwargs()` to both the sync and the async engine, so each pod's real connection use is roughly twice what a single `pool_size` reading suggests:

```
per pod:  steady  = size * 2 + 1     (+1 for the new health engine)
          ceiling = (size + maxOverflow) * 2 + 1
```

The chart comments did not reflect this. They are corrected here, and the gateway defaults are raised so a single pod is not capped at a handful of connections while the database sits idle.

**2. The pod died even though liveness is already DB-free.**
Liveness is `/api/v1/ping`, which touches no database, so the SIGKILL was surprising. Cause: `ping()` was declared `def`, not `async def`. FastAPI runs sync endpoints in Starlette's bounded anyio threadpool. Sync DB endpoints blocked those threads for up to the 30s pool-checkout timeout, so the DB-free liveness probe queued behind them, exceeded its timeout, and Kubernetes killed a pod that was merely busy. This is the actual restart mechanism, and it is a one-word fix.

## Changes

- **`websocket_manager._sync_batch_insert_logs`**: bounded retry (3 attempts, 0.5s/1.0s backoff) on `TimeoutError`/`OperationalError` before dropping; rolls back on failure so no dirty transaction returns to the pool; returns a success bool; non-retryable errors fail fast. Still notifies admins on real loss. A semaphore bounds background log persistence so it cannot starve request-serving connections. DB access stays behind `preloop.models.crud`.
- **`health.py`**: `/health` uses a dedicated 1-connection engine (`get_health_engine`) with 3s timeouts, so readiness reflects DB reachability rather than pool saturation. Previously a saturated pool marked every pod NotReady at once.
- **`health.py`**: `/ping` is now `async` (see root cause 2).
- **`values.yaml`**: chart-default pool sizing plus corrected documentation of the two-engines-per-pod arithmetic.
- **Broadcast log noise**: zero-listener broadcasts accounted for roughly 69% of gateway log lines in a sample. Downgraded to DEBUG; broadcasts with real listeners still log at INFO. This is log/CPU waste only, it performs no DB work and did not contribute to the incident.

## Deployment note

Chart defaults alone do not govern every deployment. Any environment that sets `DATABASE_POOL_SIZE` / `DATABASE_MAX_OVERFLOW` explicitly (via its own values file, deployment env, or `--set` at deploy time) must be updated separately, or this chart change has no effect there. Operators should size with the `steady`/`ceiling` formula above against their own `max_connections`, keeping steady-state totals comfortably below the usable limit.

## Tests

11 new tests: retry-then-succeed, drop-only-after-max-attempts (asserting the exact 0.5s/1.0s backoff schedule), `OperationalError` retry, non-retryable fail-fast, rollback-on-failure, empty batch opens no connection, semaphore bound, both broadcast logging levels, and two liveness regression guards.

```
backend/tests/services/test_websocket_manager.py  29 passed
backend/tests/endpoints/test_health.py             8 passed
```

Wider suite `backend/tests/{services,endpoints,api}`: **3061 passed** vs **3050 passed** on clean `main`, with the same **20 pre-existing failures** before and after (`test_openai_gateway_attribution_sessions.py` and others, unrelated). ruff + ruff-format clean; mypy shows no new errors on touched files (24 before, 24 after; the new `get_health_engine` is fully typed).

## What I did NOT change

- No cluster-side changes of any kind. No `kubectl apply/edit/scale/delete/restart`, no DB writes.
- No PostgreSQL/CNPG change. `max_connections` was not the binding constraint.
- Did not touch HPA settings, resource limits, probe timings, or the 30s `pool_timeout`.
- Did not change the app-level defaults in `config.py` (20/40), which suit single-pod and dev use; the chart governs deployed environments.
- Did not restructure the sync+async dual-engine design, which is the deeper reason per-pod connection use is 2x what it appears. Worth a follow-up.
- Author: d-mo

> [!NOTE]
> **Low Risk**
> Surgical incident response with comprehensive test coverage. No security-sensitive changes.
>
> **Overview**
> Fixes per-pod SQLAlchemy pool exhaustion through retry logic, dedicated health-check engine, async liveness probe, and corrected Helm pool sizing. Addresses two distinct root causes with 14 tests covering retry schedules, concurrency, and regression guards.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 0f8f27dc. Updates automatically on new commits.</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Database connection pool exhaustion and execution-log data loss**: a
+  production gateway pod exhausted its SQLAlchemy pool
+  (`QueuePool limit of size 3 overflow 7 reached`) under PR-reviewer load,
+  which dropped a batch of NATS execution logs, failed the readiness probe,
+  broke token validation, and ended in a pod restart. Four changes:
+  - `_sync_batch_insert_logs` now retries transient failures
+    (`TimeoutError`/`OperationalError`) up to 3 times with exponential backoff
+    (0.5s, 1.0s) before dropping a batch, rolls back on failure so no dirty
+    transaction is returned to the pool, and returns a success boolean.
+    Background log persistence is additionally bounded by a semaphore so it
+    cannot starve request-serving connections. Batches are still only dropped
+    as a last resort, and admins are still notified when that happens.
+  - Health checks use a dedicated single-connection engine with fast timeouts
+    instead of a pooled request session, so readiness reports "can I reach
+    Postgres" rather than "is the request pool momentarily full".
+  - `/api/v1/ping` (the liveness probe) is now `async`, keeping it on the event
+    loop. As a sync endpoint it ran in Starlette's bounded anyio threadpool and
+    could queue behind blocked database calls, causing Kubernetes to SIGKILL a
+    pod that was merely busy.
+  - Gateway connection pool sizing raised from 3+7 to 6+14 per pod, api tuned
+    to 8+12 and workers to 2+4. Chart comments now document that each pod
+    creates two pools (sync + async engine) and reflect real production replica
+    counts (api=2, gateway=5, workers=8).
+
+- **Gateway log noise**: WebSocket broadcasts with no matching listeners logged
+  at INFO on every event, accounting for ~69% of gateway log lines (8170 of
+  11810 in a two-hour sample). These now log at DEBUG; broadcasts with actual
+  listeners still log at INFO.
+
 ## [0.14.0] - 2026-08-07
 
 Highlights: **Cursor usage import** brings bundled-model spend into Cost
--- a/backend/preloop/api/endpoints/health.py
+++ b/backend/preloop/api/endpoints/health.py
@@ -4,22 +4,25 @@ import logging
 from datetime import UTC, datetime
 from typing import Any, Dict
 
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter
 from sqlalchemy import text
-from sqlalchemy.orm import Session
 
-from preloop.models.db.session import get_db_session as get_db
+from preloop.models.db.session import get_health_engine
 
 router = APIRouter()
 logger = logging.getLogger(__name__)
 
 
 @router.get("/ping")
-def ping() -> Dict[str, str]:
+async def ping() -> Dict[str, str]:
     """Simple liveness check - no database, no external dependencies.
 
-    Use this for Kubernetes liveness probes to avoid killing pods
-    due to temporary database issues.
+    Declared ``async`` on purpose: a plain ``def`` endpoint is executed in
+    Starlette's bounded anyio worker threadpool (40 threads by default). During
+    a database stall, sync endpoints block those threads waiting on pool
+    checkout, so even this DB-free probe would queue behind them and the
+    liveness probe would time out and get the pod SIGKILLed. Running on the
+    event loop keeps liveness answerable no matter how saturated the DB is.
 
     Returns:
         Simple pong response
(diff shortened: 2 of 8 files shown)


d-mo:
Addressed all three review items in 0f8f27dc.

**1. MEDIUM — test hygiene (live admin notifications)**
`test_sync_batch_insert_logs_database_error` now patches `preloop.services.websocket_manager.notify_admins` (matching the existing style in the file) and asserts it was called. I confirmed the gap was real, not theoretical: with `TESTING` unset, that test reached the live `notify_admins` and logged `Notifying admins: [Preloop Alert] NATS Log Persistence Failed`.

I audited the whole file with a temporary tripwire fixture. Two details mattered for making the audit trustworthy:
- it had to patch the name `websocket_manager` actually calls (the module imports `notify_admins` directly, so patching `preloop.sync.tasks.notify_admins` catches nothing);
- it had to *record* rather than `raise`, because both prod call sites wrap the alert in `try/except` and would swallow the assertion.

That surfaced a second leak I would otherwise have missed: `test_nats_consumer_handles_invalid_json` reached the live `notify_admins` through the malformed-NATS-message alert (dispatched via `asyncio.to_thread`, so it lands after the test body). It now patches and asserts too, with a short sleep for the worker thread. Re-running the file under the tripwire shows zero leaks; the tripwire was validated against a deliberately unpatched canary first. No test depends on `TESTING=true` for this.

**2. LOW — `get_health_engine` race**
Guarded with a module-level `threading.Lock` using double-checked locking: lock-free fast path once the engine exists, re-check inside the lock before creating. Comment explains why. Added a red/green test (`test_get_health_engine_is_created_once_under_concurrency`) that races 8 threads through a slowed `create_engine` — it built 8 engines before the fix, 1 after.

**3. LOW — `broadcast_json` scan**
Removed the unconditional full scan. High-frequency types (`agent_log_line`, `token_usage_update`, `tool_calls_update`) only ever log at DEBUG, so the scan is now skipped entirely behind `logger.isEnabledFor(logging.DEBUG)` — in production, where DEBUG is off, the hot path does no scan at all. Low-volume types do a single scan that feeds both the INFO and the "no listeners" DEBUG branch. The count moved into `_count_account_connections`.

I deliberately did **not** maintain a per-account counter. It would mean new bookkeeping in `connect_with_account`/`disconnect` (plus the `active_connections`-only `connect()` path, which never registers an account), and a counter that drifts out of sync silently produces wrong log lines. The existence/lazy-scan fix removes the cost from the hot path with no changes to connection bookkeeping. Two tests pin the behaviour: one asserts no scan when DEBUG is off (via a dict that counts `.values()` calls), one asserts the count is still accurate when DEBUG is on.

**Tests**
- `tests/services/test_websocket_manager.py`: 31 passed (29 + 2 new)
- `tests/endpoints/test_health.py`: 8 passed
- `tests/models/test_db_session.py`: 10 passed (+1 new)
- `ruff format` / `ruff check` clean; `mypy` on both changed source files produces a byte-identical error set to the pre-change baseline (the file has pre-existing untyped-def errors; I added none).
- Wider run of `tests/services/` + `tests/endpoints/test_websockets.py`: 2162 passed / 20 failed, and the same 20 fail on the parent commit (unrelated, pre-existing — they need a live DB).

PR #197, opened 2026-08-08
fix: price OpenRouter-routed models and stop reporting unpriced usage as $0

d-mo:
## The bug

Flow executions were displaying `estimated_cost = 0.0000` even though the gateway had correctly metered **62.9M tokens** across 1,667 calls on the affected models. At the provider's published prices that traffic has a real, non-zero cost. We were reporting a confident $0.00 for traffic we simply could not price.

Reporting a confident zero is worse than reporting nothing: cost credibility is the core value proposition.

## Root cause

Three defects compounded. Token metering was never broken; pricing and presentation were.

1. **Alias never matched the catalog.** Usage on a user-configured OpenAI-compatible provider is recorded as `openai-compatible/<vendor>/<model>`. `openai-compatible` is a Preloop *routing* label, not a price-map namespace, and it was carried straight into catalog lookup, guaranteeing a miss. OpenRouter-routed models were also never tried under litellm's `openrouter/vendor/model` key form.
2. **No price source for the model.** litellm's map does not carry dated OpenRouter SKUs at all, so even a correctly-formed key missed. `cost_source` was set to `unpriced` and `estimated_cost` left NULL, which is correct behavior.
3. **The NULL was then rendered as a confident zero.** `get_gateway_usage_for_execution` wrapped `sum(estimated_cost)` in `coalesce(..., 0.0)`, and `FlowOrchestrator.estimated_cost` defaulted to `0.0`. "We could not price this" was persisted and displayed as "this was free". The frontend compounded it by treating a numeric `0` cost as proof of pricing.

## A trap worth calling out

The obvious fix, stripping the `-0731` date suffix to reach the undated entry, would have been **wrong**. Verified against OpenRouter's API:

| model | prompt /Mtok | completion /Mtok |
|---|---|---|
| `deepseek-v4-flash-0731` (dated, in use) | **$0.09** | **$0.18** |
| `deepseek-v4-flash` (undated) | $0.14 | $0.28 |

Dated marketplace ids are separately priced SKUs. That fallback would have overstated the bill by ~55%. Date-suffix stripping is now explicitly disabled for `vendor/model` names. A wrong number in a cost dashboard is worse than an honest "unknown".

## Platform-wide unpriced volume

`api_usage` grouped by `cost_source`:

| cost_source | calls | tokens |
|---|---|---|
| **unpriced** | **1,727** | **63,005,680** |
| subscription (legitimate $0) | 13,173 | 13,995,362 |
| catalog (priced) | 53 | 339,846 |
| null (mostly non-token rows) | 490,122 | 2,165,591 |

**Essentially all metered token volume on the platform, ~63M tokens, was unpriced**, concentrated in OpenRouter-routed models. Other affected models: `moonshotai/kimi-k3`, `openrouter/auto-beta`, `kimi-for-coding/*`, several `gpt-5-codex` variants.

## What is fixed

- **Alias normalization**: synthetic `openai-compatible/`, `custom/`, `preloop/` prefixes stripped before lookup; models served from `openrouter.ai` also tried under `openrouter/vendor/model`.
- **Real price source**: OpenRouter's own `/api/v1/models` is used when litellm's map misses, cached and backed off like the existing price-map fetch. Zero/missing vendor prices are skipped rather than registered as "free". Verified end to end: an affected dated SKU now resolves to its exact per-Mtok input/output price from `catalog`.
- **Never a fake zero** (founder directive A): cost stays `NULL` when nothing could be priced and the **token count is surfaced in its place**. Aggregates mixing priced and unpriced rows now expose `cost_is_partial` plus unpriced request/token counts, so a subtotal is never presented as a complete bill. Frontend shows `cost unavailable: model not priced` instead of `$0.00`.
- **Subscription zero preserved**: `cost_source='subscription'` is a legitimate, complete $0.00 and is explicitly excluded from unpriced counting.
- **Admin alert** (founder directive B): first time a `(model_alias, provider)` proves unpriceable, `notify_admins` fires with the model, account id and token volume. Deduped via a **persisted `audit_log` marker** with a 24h cooldown, so it holds across the multiple prod replicas rather than firing once per process (the failure mode of a past incident). The cooldown is claimed *before* notifying, and every DB/notification error is swallowed: alerting can never break a user request. 1,667 unpriced calls produce ~1 alert.

## Why our own alarm never fired (systemic)

The `costs_priced` check in `gateway_accounting_check.py` works correctly and would have failed on this traffic. It is simply **never invoked**: it is exposed only as `GET /api/v1/cost/health`, is per-account, requires auth, and has no scheduler, cron or k8s job polling it. Nobody was ever going to be told. The new push-based alert on the recording path closes that gap; wiring the health check into a scheduled sweep is a sensible follow-up.

## Zero-token executions (investigated, not ours)

28 of 91 executions recorded zero tokens. All 28 are `status = FAILED` **with zero `api_usage` rows**, failing at `Fetching merge request ref pull/N/head... FATAL` before any model call. This is lost git/checkout work, not lost accounting, and is unrelated to the DB pool incident. No fix here; it belongs with the PR-reviewer reliability work.

## Founder actions required

**1. Backfill (prepared, NOT executed).** Nothing was run against prod. Dry run first:

```bash
python scripts/reprice_unpriced_usage.py --all-accounts --days 30
python scripts/reprice_unpriced_usage.py --all-accounts --days 30 --apply
```

Scope: **1,684 rows** across **2 accounts**, 2026-07-26 to 2026-08-08. Repricing is analytics-only: budget spend is never rewritten and subscription rows are skipped.

**2. Prod egress.** The gateway pods must be able to reach `https://openrouter.ai/api/v1/models` for live pricing. If egress is restricted, allowlist it or set `OPENROUTER_MODELS_URL` to a mirror. Without it, affected models stay honestly unpriced (tokens shown) rather than falsely $0.00.

**3. Optional.** `UNPRICED_MODEL_ALERT_COOLDOWN_HOURS` (default 24) tunes alert frequency.

## Tests

Red-first throughout. Added coverage for alias normalization on openai-compatible/OpenRouter forms, the dated-SKU guard, OpenRouter price conversion (including refusing zero prices), unpriced never rendering as 0 and rendering tokens instead, mixed aggregates flagging the unpriced remainder, subscription-zero staying a legitimate 0, and `notify_admins` firing exactly once per (model, provider) per cooldown, surviving a simulated replica restart, and never raising into the request path.

`test_model_pricing.py`, `test_model_price_catalog.py`, `test_usage_repricing.py`, `test_cost_health.py`, `test_execution_metrics.py`, `test_unpriced_model_alert.py`, `test_unpriced_usage_reporting.py`, `test_replay_usage_exclusion.py`: **90 passed**. Full `tests/endpoints`: **865 passed**. Frontend `flow-execution-view`: **6 passed**. Pre-existing unrelated failures (`test_openai_gateway_attribution_sessions.py`, `test_ai_model_updated_at_trigger.py`, gateway e2e) were confirmed identical on clean `main` before and after. ruff format + check clean; mypy on the new module clean.

- Author: d-mo

> [!NOTE]
> **Low Risk** — ✅ Approved
> Thorough bug fix with defensive patterns, comprehensive tests, and no security concerns. The change to cost display semantics (NULL vs 0.0) is the most impactful user-facing shift and is well-tested.
>
> **Overview**
> Fixes three compounding defects that caused OpenRouter-routed and other unpriceable gateway traffic to display a confident $0.00 cost. Strips synthetic routing prefixes before pricing lookup, adds OpenRouter's own `/api/v1/models` as a secondary price source, and ensures NULL costs are never coalesced to zero. Adds cross-replica admin alerting for unpriceable models via persisted audit_log markers.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 757d708. Updates automatically on new commits.</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,43 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   remote plus available refs, so the log shows whether the commit was
   force-pushed away, the ref is missing, or credentials failed.
 
+- **Unpriced usage no longer reports as $0.00**: gateway traffic routed through
+  OpenRouter/`openai-compatible` endpoints was metered correctly (tokens
+  captured) but could never be priced, so flows that cost real money displayed
+  a confident `$0.00`. Three defects combined: the synthetic
+  `openai-compatible/` prefix was carried into price-catalog lookups where it
+  can never match; OpenRouter-routed models were not tried under litellm's
+  `openrouter/vendor/model` keys; and a `sum()` over NULL costs was coalesced
+  to `0.0` in `get_gateway_usage_for_execution`, with `FlowOrchestrator`
+  defaulting `estimated_cost` to `0.0`. Cost now stays NULL when nothing could
+  be priced, and the token volume is surfaced instead of a fake zero. Aggregates
+  that mix priced and unpriced rows expose `cost_is_partial` plus the unpriced
+  request/token counts so a subtotal is never presented as a complete bill.
+  Subscription-covered traffic (`cost_source='subscription'`) is unchanged and
+  still reports a legitimate `$0.00`.
+
+- **OpenRouter model pricing**: models served from `openrouter.ai` are now
+  priced from OpenRouter's own `/api/v1/models` endpoint when litellm's map
+  does not carry them, cached and backed off like the existing price-map fetch.
+  Date-stamped marketplace ids (e.g. `deepseek-v4-flash-0731`) are deliberately
+  NOT aliased to their undated entry: they are separately priced SKUs, and the
+  fallback would have overstated cost by ~55% for that model. Models that still
+  cannot be priced stay explicitly unpriced rather than being given a guess.
+
+### Added
+
+- **Admin alert for unpriceable models**: the gateway now notifies admins the
+  first time a `(model_alias, provider)` pair proves unpriceable, including the
+  account and token volume, so missing pricing is noticed instead of silently
+  surfacing as no spend. Deduplicated via a persisted `audit_log` marker with a
+  24h cooldown (`UNPRICED_MODEL_ALERT_COOLDOWN_HOURS`), so it holds across
+  replicas rather than firing once per process, and every failure path is
+  swallowed so alerting can never break a user request.
+
+- **`scripts/reprice_unpriced_usage.py`**: operator script to backfill costs for
+  historical rows recorded while a model was unpriced. Dry run by default;
+  requires `--apply` to persist.
+
 ## [0.14.0] - 2026-08-07
 
 Highlights: **Cursor usage import** brings bundled-model spend into Cost
--- a/backend/preloop/api/endpoints/flows.py
+++ b/backend/preloop/api/endpoints/flows.py
@@ -579,8 +579,13 @@ def get_flow_execution_metrics(
         - tool_calls: Number of MCP tool calls made
         - api_requests: Number of API requests made during execution
         - token_usage: Token usage statistics
-        - estimated_cost: Estimated cost based on token usage (0.0 if no pricing)
-        - has_pricing: Whether pricing is configured in AI model metadata
+        - estimated_cost: Estimated cost in USD, or null when no usage could
+          be priced (never a placeholder 0.0, which would misreport real
+          spend as free)
+        - has_pricing: Whether any request could be priced
+        - cost_is_partial: Whether estimated_cost excludes unpriced requests
+        - unpriced_requests: Requests that could not be priced
+        - unpriced_tokens: Token volume behind the unpriced requests
     """
     from preloop.services.execution_metrics import ExecutionMetricsService
 
@@ -606,8 +611,13 @@ def get_flow_execution_metrics(
             "tool_calls": 0,
             "api_requests": 0,
             "token_usage": {"total_tokens": 0, "input_tokens": 0, "output_tokens": 0},
-            "estimated_cost": 0.0,
+            # Metrics could not be computed, so the cost is unknown rather
+            # than zero; the UI renders this as unavailable, not as free.
+            "estimated_cost": None,
             "has_pricing": False,
+            "cost_is_partial": False,
+            "unpriced_requests": 0,
+            "unpriced_tokens": 0,
         }
 
 
--- a/backend/preloop/models/crud/api_usage.py
+++ b/backend/preloop/models/crud/api_usage.py
@@ -1498,7 +1498,26 @@ class CRUDApiUsage(CRUDBase[ApiUsage]):
         db: Session,
         execution_id: str,
     ) -> Dict[str, Any]:
-        """Return explicit gateway usage totals for an execution when available."""
+        """Return explicit gateway usage totals for an execution when available.
+
+        ``estimated_cost`` is deliberately NOT coalesced to ``0.0``: a NULL
+        cost means "we could not price this", which is different from "this
+        was free". When no request could be priced the cost stays ``None`` and
+        callers render the token volume instead (see ``unpriced_tokens``).
+
+        Args:
+            db: Database session.
+            execution_id: Owning flow execution id.
+
+        Returns:
+            Usage totals including ``estimated_cost`` (``None`` when nothing
+            was priceable), ``cost_is_partial`` when priced and unpriced rows
+            are mixed, and the unpriced request/token volume.
+        """
+        unpriced_condition = and_(
+            ApiUsage.estimated_cost.is_(None),
+            func.coalesce(ApiUsage.cost_source, "") != "subscription",
+        )
         row = (
             db.query(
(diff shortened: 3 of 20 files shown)


d-mo:
Rebased onto `main`, addressed the code-quality feedback, and folded in the Dependabot fixes. Three commits now, deliberately kept separate.

## 1. Rebase onto `main` (13 commits, PRs #194/#195/#196)

Three files were touched on both sides. Only `CHANGELOG.md` actually conflicted; the other two auto-merged, and I verified both semantically rather than trusting the merge.

**`flow_orchestrator.py` — the one that mattered.** This was the real semantic risk: #196 changed flow-execution code (loop-guard false positive, stop status, checkout diagnostics) in the same file where this branch changes `estimated_cost` to stay `None`. Git auto-merged it, so I checked what that actually produced by diffing the rebased file against `github/main`:

```
$ git diff github/main -- backend/preloop/services/flow_orchestrator.py
-        self.estimated_cost: float = 0.0
+        # None means "not priced yet / could not be priced". Defaulting this to
+        # 0.0 made unpriced executions display a confident $0.00 in the UI.
+        self.estimated_cost: Optional[float] = None
```

That single hunk is the *entire* delta against `main` — so all of #196's reliability work is intact and the `0.0` default did not sneak back in. Confirmed positively too: `STOPPED` status, the "repeated MCP tool loop" guard and the checkout diagnostics are all still present in the file, and #196's own regression tests (`test_mcp.py`, `test_container.py`, `test_flow_orchestrator_streaming.py`) pass. Verified by test rather than by eye, as requested.

- `CHANGELOG.md`: kept both sets of entries under `Unreleased`.
- `openapi.yaml`: only this branch's `cost_is_partial` / unpriced-count additions remain versus `main`; re-parsed with `yaml.safe_load` (valid, 162 paths).

## 2. `_PriceMapCache` refactor (github-code-quality bot)

The bot flagged `_openrouter_map` / `_openrouter_fetched_at` / `_openrouter_failed_at` as unused globals (6 comments). That was a false positive — they were read, written and reset, but the analyzer does not follow module-level `global` rebinding. Deleting them would have broken the cache and the failure backoff, so I removed the **underlying smell** instead, which was real: `_fetch_remote_price_map` and `_fetch_openrouter_price_map` were near-identical "cached fetch with TTL and failure backoff" implementations over two parallel triples of module globals, each rebinding from four separate places.

Both now share one small `_PriceMapCache` holding `(map, fetched_at, failed_at)` with `get` / `set` / `mark_failed` / `reset` / `snapshot`, guarded by the existing `_lookup_lock` and instantiated twice. The download functions became pure "fetch or return `None`"; the cache alone decides TTL reuse, backoff and storage.

Behaviour is deliberately unchanged: same TTL and backoff windows, same retry count and SHA-256 pinning for litellm's map, same "empty entries counts as a failure" rule for OpenRouter, same negative cache, same "both sources unavailable → return `None` rather than negative-cache", same lock, same reset-for-tests semantics. The module-level `global` rebinding is gone entirely, so the warning disappears for the right reason rather than by suppression. Added tests for TTL reuse/expiry, backoff and its expiry, OpenRouter backoff parity, and reset clearing both caches.

## 3. Dependency security updates (separate commit)

All four open Dependabot alerts were npm/frontend and cleared with in-range patch bumps — no major jumps needed. `js-yaml` 4.3.0 → 4.3.1 (high, but build-time only: the sole importer is `vite-plugin-brand.ts` parsing in-repo config, so untrusted YAML never reaches it) and `mermaid` 10.9.6 → 10.9.8 (three alerts; declared in `package.json` but not imported anywhere, so the vulnerable code never ships in a bundle). Also picked up `nanoid` 3.3.16 → 3.3.17 transitively — lockfile only, so it does not become a direct dependency.

Deferred: the remaining `npm audit` findings are all in the `deck.gl` / `@loaders.gl` / `@luma.gl` tree with no non-breaking fix (`npm audit fix --force` wants a major `deck.gl` downgrade). None are open Dependabot alerts. Full table is in the PR description.

Kept as its own commit on top of the cost work so it can be dropped or reverted independently — the cost fix is the customer-facing priority and should not be held hostage by a dependency upgrade.

## Tests

Backend `tests/services` + `tests/models`: **2796 passed, 23 failed**. Baselined against clean `github/main` in a scratch worktree: **2777 passed, 23 failed** — byte-identical failure set (`test_openai_gateway_attribution_sessions.py` ×20, `test_ai_model_updated_at_trigger.py` ×3, both known pre-existing). **Zero regressions**, and 19 net new passing tests.

Targeted suites (`test_model_price_catalog`, `test_model_pricing`, `test_unpriced_model_alert`, `test_execution_metrics`, `test_usage_repricing`, `test_unpriced_usage_reporting`, `test_flow_orchestrator_streaming`): **101 passed**. Frontend after the bumps: `npm ci`, `npm run build`, `npm test` (85 files, **618 passed**) all clean. `ruff format` + `ruff check` clean; `mypy` on `model_price_catalog.py` reports **0 errors**.

PR #205, opened 2026-08-09
feat: show flow execution duration beside the start time

d-mo:
## Why

The flow executions table showed **Start Time** next to **End Time**. Between
them they say when a run happened and when it stopped, but neither answers the
question people actually scan that table for: *how long did it take?* Working
that out meant subtracting two wall-clock timestamps by eye, and for a run
still in flight there was nothing at all — the End Time cell was just `-`.

Both timestamps have always been on the API (`start_time` non-null,
`end_time` nullable), so this is a console-only change.

## What changed

**New formatter** — `formatDurationBetween(start, end?, now?)` in
`utils/date.ts`. It tolerates the messy cases the UI actually hits:

- no end time → measures elapsed against `now` (injectable, so tests are
  deterministic)
- missing/unparseable start, garbage end, or an end before the start
  (clock skew) → returns `''` rather than `NaN` or a negative span, letting
  each caller pick its own placeholder

`calculateDuration()` now delegates to it, so the existing output style
(`1h 5m` / `4m 32s` / `25s`) is shared and unchanged.

**Rendering rule**, applied consistently:

| State | Shown |
|---|---|
| `end_time` present | the finished span, e.g. `4m 32s` |
| No `end_time`, still running | `Running · 4m 32s` |
| No `end_time`, terminal status (legacy rows) | `—` in tables, nothing inline |
| Unusable timestamps | `—` in tables, nothing inline |

That last row matters: those legacy rows previously rendered `Running...` on
the flow detail page, claiming a long-dead execution was still going.

**Where:** executions table (End Time column → Duration), flow detail recent
executions table, Flows list and dashboard rows (appended to the
`Started …` line), and the execution detail Timing card.

**Live ticking** is limited to the detail view, where a single interval
advances a `durationNow` state once a second. It stops when the status turns
terminal and on `disconnectedCallback`, so nothing is left re-rendering a
finished run. List views compute elapsed time at render and add **no** timers
— rows already refresh via the existing polling/WebSocket updates.

## Testing

Written red-first, then implemented.

- `date.test.ts` — 12 cases on the formatter: each format bucket, ≥24h staying
  in hours, naive backend strings as UTC, missing/reversed/garbage inputs, an
  explicit `now`, plus a `calculateDuration` regression check.
- `flow-executions-view.test.ts` — header is `Duration` and no longer
  `End Time`; completed / running / legacy-terminal / clock-skewed rows.
- `flow-execution-view.test.ts` — completed and running Timing card, the value
  advancing under fake timers, and the interval cleared on disconnect.
- `flows-view.test.ts` — duration appended for finished and running runs,
  nothing appended for a legacy terminal row.

Full suite: **86 files, 648 passed, 0 failed**. `tsc --noEmit` reports the same
81 pre-existing errors as `main` — none new. No backend tests: no API change.

## Out of scope

Backend/OpenAPI (nothing to add), Duration sorting/filtering, per-row tickers
in lists, and runtime-session durations.

> [!NOTE]
> **Low Risk**
> Well-tested frontend-only change with thorough edge-case handling and no API modifications.
>
> **Overview**
> Replaces raw end timestamps with computed durations across the console UI. Adds a `formatDurationBetween` utility that gracefully handles missing/garbage/clock-skewed timestamps. Live ticking is scoped to the execution detail view only. All previous review feedback has been addressed — duplicated constants and methods are now extracted to a shared `utils/execution.ts` module.
>
> <sup>Written by Preloop PR Reviewer for commit `54c2fa2`. Updates automatically on new commits.</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -157,6 +157,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Flow execution duration in the console**: the executions table now shows a
+  **Duration** column in place of the raw "End Time" (the start time already
+  said when the run happened; the end time alone never said how long it took),
+  and the same value is appended to the "Started …" line on the Flows and
+  dashboard execution lists. Running executions display `Running · <elapsed>`,
+  ticking every second on the execution detail page and recomputed on each
+  render elsewhere; runs that ended without an `end_time` show `—` instead of
+  claiming to still be running. Both timestamps were already returned by the
+  API, so this is a console-only change.
+
 - **Admin alert for unpriceable models**: the gateway now notifies admins the
   first time a `(model_alias, provider)` pair proves unpriceable, including the
   account and token volume, so missing pricing is noticed instead of silently
--- a/frontend/src/utils/date.test.ts
+++ b/frontend/src/utils/date.test.ts
@@ -1,5 +1,7 @@
 import { expect } from '@open-wc/testing';
 import {
+  calculateDuration,
+  formatDurationBetween,
   formatFutureRelativeTime,
   formatRelativeTime,
   parseUTCDate,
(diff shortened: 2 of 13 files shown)


d-mo:
Addressed the review in 54c2fa26:

- **Duplicated `RUNNING_STATUSES` (medium)** — fixed. Extracted to `frontend/src/utils/execution.ts` as a single exported `ReadonlySet`; all four views now import it.
- **Duplicated `executionDuration` method (medium)** — fixed. Replaced the four copies with a shared `executionDurationText(exec, now?)` in the same module, with its own unit tests. The detail view's Timing card also delegates to it (passing its ticking `durationNow`), so there is now exactly one implementation of the rendering rule.
- **`isExecutionRunning` inline `||` comparisons (low)** — fixed. It now checks membership in the shared `RUNNING_STATUSES`.

No behaviour change: all previously-passing view tests still pin the rendered output (finished span, `Running · <elapsed>`, em dash for legacy/unusable rows). Full frontend suite green (87 test files, 654 passed), pre-commit hooks pass, and `tsc --noEmit` shows the same 123 pre-existing errors as before the change (none new).

PR #207, opened 2026-08-09
PR Reviewer preset: bounded reads + incremental re-review; link renamed preset clones by content hash

d-mo:
## Summary

Two independent commits, one theme: preset flows should be cheap to run and easy to keep up to date.

### 1. Token-optimised PR Reviewer preset

The stock Pull Request Reviewer prompt let agents read without bounds: full CHANGELOG (~40k tokens on a mature repo), open-ended "understand the codebase" repo walks, whole-file opens for two-line hunks, and unbounded verification searches. Fresh reads during the run are exactly the part of a review that prompt caching does not cover, so this is where the bill is.

Changes (all hard rules — agent CLIs follow hard bounds well):

- **Bounded doc reads**: agent-instruction files in full; README/ARCHITECTURE/CONTRIBUTING capped at the first ~150 lines; CHANGELOG dropped entirely (head −30 only when the diff touches it); manifests/CI/linter configs read only when the diff touches deps/CI/style.
- **Bounded context discovery**: language/framework/conventions derived from the diff + at most 3 files outside it. No repo walking.
- **Hunk-first reading**: work from diff hunks; open surrounding context (enclosing function/class, not whole files) only when a finding needs it.
- **Per-finding verification budget**: the five verification checks stay intact, but capped at 2 greps + 2 file reads per finding; unresolved findings get filed as questions instead of triggering search spirals.
- **Conditional documentation-impact pass**: runs only when the diff adds user-facing surface (endpoints, config, env vars, CLI, features).
- **Region-scoped re-verification**: previous findings are re-checked against the ±40-line region or named symbol, not the whole file. Quote-it-or-drop-it unchanged.
- **Skip identical PR-description rewrites**; **replace status stamps instead of stacking them**, and only touch comments whose status changed.
- **Single-fetch rule**: `get_pull_request` exactly once; everything git-related comes from the clone.
- **Small-PR fast path**: <~50 changed lines and no previous Preloop comments → skip discussion scan, doc reads, and doc-impact; the security/quality checks still run on the same hunks.
- **Leaner output**: empty severity sections omitted; "What Looks Good" dropped when requesting changes. The Affected-files block is kept (load-bearing for cross-flow use).

**New: incremental re-review.** The summary comment now records the reviewed HEAD SHA in a `<!-- preloop-review:reviewed-sha:... -->` marker. When the trigger is `pull_request_updated` and a marker exists, the reviewer validates it in the clone (`git cat-file -e` + `git merge-base --is-ancestor` to detect force-push/rebase), reviews only `git diff <sha>..HEAD` hunks, and re-verifies only previous findings whose files changed — with an explicit full-review fallback whenever the delta cannot be derived safely. **Expected effect: per-push review cost becomes proportional to the push delta instead of the whole PR.**

Not included: a cap on discussion replies (audit finding A7) — pending a product decision.

Prompt length: 799 lines (was 834).

### 2. Fix: preset updates never reached renamed flow clones

`sync_preset_to_derived_flows` finds derived flows via `source_preset_id`, but the one-time linking migration only matched flows named `Copy of <preset name>`. Operators who cloned a preset and renamed the flow — prompt still byte-identical to the preset — ended up with `source_preset_id=NULL`, and preset updates silently never reached a customer's flows.

New `link_unlinked_flows_by_content` pass (service + wired into `scripts/sync_flow_presets.py`, both in `--link-existing` and before propagation in the default run) links unlinked, non-preset, account-owned flows whose prompt hash equals a preset's current prompt **or** a historical link-time version of it (recovered from the distinct `source_prompt_hash` values of already-linked flows), regardless of name.

Conservative by construction:
- Only byte-identical prompts link — a customized prompt can never match, so user edits can never be overwritten by this pass.
- A hash matching multiple presets is skipped and logged; no guessing.
- Flags are set exactly like the name-based path; differing tools are marked `tools_customized` + `preset_update_available` (notified, never replaced).

## Test plan

- [x] `tests/test_flow_presets.py` (14 passed) — preset schema/loading, revised YAML parses
- [x] `tests/services/test_preset_content_hash_linking.py` (6 passed, new) — renamed identical clone linked then auto-updated; customized flow never linked/touched; older-preset-version clone linked at old hash and updated on next sync; ambiguous hash skipped; customized tools notified not overwritten; dry-run mutates nothing
- [x] `tests/services/test_flow_presets_service.py` (55 passed) — existing sync semantics unchanged
- [x] pre-commit (ruff, ruff-format, yaml check) clean

> [!NOTE]
> **[Low Risk]**
> No security concerns. The content-hash linking is conservative (byte-identical only) and the preset prompt changes are instruction-only with no code execution changes.

> **Overview**
> Two independent improvements: a token-optimised PR Reviewer preset prompt with incremental re-review (bounds reads, adds verification budgets, and supports per-push delta reviews), and a content-hash based linking pass that fixes preset updates silently failing to reach renamed-but-unmodified cloned flows. Well-tested with 6 new test cases and 75 existing tests unchanged. Previously noted N+1 query resolved in this revision.

> <sup>Written by [Preloop PR Reviewer](https://preloop.ai). Updates automatically on new commits.</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **PR Reviewer preset: token-optimised prompt**: the stock Pull Request
+  Reviewer preset now bounds every open-ended read that previously let agents
+  walk the repository. Project doc reads are capped (agent-instruction files in
+  full, README/ARCHITECTURE/CONTRIBUTING heads only, CHANGELOG dropped,
+  manifests/CI/linter configs only when the diff touches them); project-context
+  discovery is limited to the diff plus at most 3 files outside it; Phase 2
+  works hunk-first instead of opening whole changed files; each finding gets a
+  verification budget (2 greps + 2 file reads, then phrase as a question); the
+  documentation-impact pass runs only when the diff adds user-facing surface;
+  previous-finding re-verification reads the ±40-line region instead of the
+  whole file; the PR description is only rewritten when its content changed;
+  persisting-issue stamps are replaced instead of stacked and unchanged-status
+  comments are left alone; a single-fetch rule forbids re-calling
+  `get_pull_request`; empty severity sections are omitted from the summary; and
+  small first-time PRs (<~50 changed lines) take a fast path that skips the
+  ceremony while keeping the full security/quality checks. New: incremental
+  re-review — the summary comment now records the reviewed HEAD SHA in a
+  `<!-- preloop-review:reviewed-sha:... -->` marker, and on
+  `pull_request_updated` triggers the reviewer diffs against that SHA via git
+  in the clone and reviews only the new hunks (with a full-review fallback on
+  force-push/rebase or a missing marker), making per-push review cost
+  proportional to the push delta instead of the whole PR.
+
 ### Fixed
 
 - **Unhelpful failure messages and no retry when an upstream model provider
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Preset updates never reached renamed flow clones**: preset propagation
+  (`sync_preset_to_derived_flows`) only finds flows via `source_preset_id`,
+  and the one-time linking migration only matched flows named
+  "Copy of <preset name>". A flow cloned from a preset and then renamed —
+  with a prompt still byte-identical to the preset — stayed unlinked forever
+  and silently never received preset updates. A new content-hash linking pass
+  (`link_unlinked_flows_by_content`, also run by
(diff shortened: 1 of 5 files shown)


d-mo:
Review feedback addressed (rebased onto main @ 344f3588, head is now cb92a0f):

- **Medium — N+1 query for historical hashes**: Fixed. The per-preset `DISTINCT source_prompt_hash` query is replaced with a single `SELECT DISTINCT source_preset_id, source_prompt_hash FROM flows WHERE source_preset_id IN (...)` over all presets, mapped back via `presets_by_id`. Behavior unchanged; all 75 preset-suite tests still pass.
- **Low — loading all unlinked account-owned flows into memory**: Intentionally left as-is. It mirrors the existing `link_existing_flows_to_presets` pattern, runs only as an operator-invoked sync/migration pass (not on a request path), and the candidate set (unlinked, non-preset, account-owned flows) is small in practice. Hash computation requires the prompt text anyway, so streaming would save little; happy to revisit with `yield_per` if the dataset ever grows.

Also resolved the CHANGELOG conflict from #205/#206 (both sides kept under `### Fixed`); no code conflicts with either PR.

Issue #212, opened 2026-08-10
PR reviewer flows fail at end of run after token-optimised preset: agent exits without success sentinel despite completed review


d-mo:
Since the token-optimised PR reviewer preset (#207) propagated to flows (2026-08-10 ~15:20 UTC), review executions complete their review (comments posted) but the execution is marked FAILED at the end: the agent exits (Gemini CLI rc=0, OpenCode rc=1) without printing FLOW_EXECUTION_SUCCESS.

## Observed (staging + prod, read-only)
- Failure rate on an affected prod flow: 0/30 before the preset update, 9/32 after (same day). Model unchanged on prod (rules out model switch as cause).
- Reviews ARE delivered; failure happens in late phases (7: update previous comments / 8: remove reaction).

## Mechanisms found in execution logs
1. Gemini CLI: burst of malformed tool calls after the update ("params must have required property 'pull_request'" / "'command'" - 46 occurrences post-update, zero before), and gemini-cli's internal LoopDetectionService/web_fetch call a hardcoded gemini-2.5-flash through the gateway -> 404 ModelNotFoundError -> turn aborts -> CLI exits 0 without the sentinel.
   Hypothesis: new Step 3.2 ("re-verify every previous issue by re-reading each file") produces repetitive read loops that trip gemini-cli's loop detector.
2. OpenCode: last chat/completions requests reach ~93k tokens, then no LLM traffic for ~2 min and "OpenCode CLI exited with code: 1" with stderr apparently not captured (blank lines in agent logs).

## Impact
- Runs marked FAILED that delivered a complete review (billing/trust: users see failed runs that worked).
- Affects any flow derived from the preset.

## Fix directions to evaluate
- Gateway: serve/alias the gemini-cli utility model (or map unknown internal helper-model requests) so loop-detection/web_fetch calls don't 404.
- Agent config: disable gemini-cli loop detection or point its helper model at the flow's model.
- Preset: bound Step 3.2's re-read pattern (batch reads, avoid per-item repetitive identical calls).
- Agent wrapper: capture OpenCode stderr so exits are diagnosable.
- Orchestrator: consider sentinel-adjacent robustness (e.g. detect submitted review + clean exit) without weakening the sentinel contract.

PR #232, opened 2026-08-16
fix(flows): stop failing exit-0 runs on missing model-printed sentinel

d-mo:
## Problem

Staging execution `ff1294e1-8e26-435d-b37f-e3a85a44b310` (internal id): the PR-reviewer agent completed all work (review approved and posted), the harness logged `OpenCode CLI exited with code: 0`, the runner emitted `agent_status SUCCEEDED` — and the platform then marked the execution FAILED with *"Agent exited with code 0 but did not produce the success sentinel"*.

The contract was one-sided: the wrapper scripts emit `PRELOOP_AGENT_EXEC_START` but no end marker, and success depended on the **model** printing `FLOW_EXECUTION_SUCCESS` — a signal models routinely skip, and which the resolved prompt in this run never even contained.

## Fix

Success decision in `_monitor_agent_execution` is now, in order of authority:

1. **result.json artifact** (`preloop.eval.result/v1`, first-class capture from #231): a `fail`/`error` status fails the run even on exit code 0 (`result_artifact_failure_override` milestone). Capture-error wrappers (`{"error": ...}`) carry no `status` and cannot flip a run.
2. **Container exit code** — unchanged, including the executor's log-based critical-error detection.
3. **Wrapper-emitted `PRELOOP_AGENT_EXEC_END:<exit_code>`** (new): printed by the same wrapper code that prints `PRELOOP_AGENT_EXEC_START`, right after the CLI exits (opencode/codex/gemini). END + exit 0 confirms clean completion — model output is never required for success. The model-printed sentinel is demoted to a legacy fallback: when it is the only signal missing, we log `sentinel_missing_warning` and keep SUCCEEDED.

Also: the post-completion grace-period path (container lingering on git push / PR creation) now arms on wrapper END:0 as well, not just the sentinel. `aider`/`openhands` are untouched — they never emitted EXEC_START, so the old override never applied to them (`_agent_exec_started` guard preserved).

Prior related work checked to avoid regressions: #212-class gemini exit-0 mitigations (helper alias pinning kept intact) and #231 artifact capture (now honored as the top-priority signal).

## Tests

- `test_exit_zero_without_sentinel_succeeds` — regression for ff1294e1 (exec start seen, no sentinel, no END marker → SUCCEEDED with warning)
- `test_exit_zero_with_wrapper_end_marker_succeeds` — END:0 → SUCCEEDED
- `test_nonzero_exit_fails` — leniency does not weaken real failure detection
- `test_result_artifact_failure_overrides_exit_zero` — artifact `status: fail` fails the run on exit 0 and is still persisted

Tests stream real log lines through the background log-streaming task (async generator + yielding fake sleep) so the markers are consumed before the status decision.

Validation: ruff check/format clean; local test DB (alembic upgrade head) — `test_flow_orchestrator.py` + `tests/agents` 329 passed / 5 skipped; matrix, retry, commit-status and flows endpoint suites 80 passed.

> [!NOTE]
> **[Low Risk]**
> The change is a well-scoped, well-tested reliability fix that removes a false-failure path without weakening real failure detection (executor critical-error detection and the artifact override are preserved). No security impact.
>
> **Overview**
> Demotes the model-printed `FLOW_EXECUTION_SUCCESS` sentinel from an exit-0 failure condition to a legacy fallback, adding a wrapper-emitted `PRELOOP_AGENT_EXEC_END:<code>` marker and honoring the result.json artifact as the authoritative signal. One medium finding: the artifact failure override is not applied on the grace-period/completion-signal path.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 5053651e. Updates automatically on new commits.</sup>

### DIFF
--- a/backend/preloop/agents/codex.py
+++ b/backend/preloop/agents/codex.py
@@ -427,6 +427,10 @@ echo "PRELOOP_AGENT_EXEC_START"
 # Run codex in non-interactive mode with the prompt
 echo "{escaped_prompt}" | codex exec --skip-git-repo-check --model "{model}" --sandbox workspace-write --yolo
 CODEX_EXIT_CODE=$?
+# Wrapper-emitted completion marker: pairs with PRELOOP_AGENT_EXEC_START.
+# The orchestrator treats this marker with exit code 0 as success without
+# requiring any model-printed sentinel.
+echo "PRELOOP_AGENT_EXEC_END:$CODEX_EXIT_CODE"
 {post_exec_block}
 # Exit with codex's exit code
 exit $CODEX_EXIT_CODE
--- a/backend/preloop/agents/gemini.py
+++ b/backend/preloop/agents/gemini.py
@@ -507,6 +507,10 @@ echo ""
 echo "=================================================="
 echo "Gemini CLI exited with code: $GEMINI_EXIT_CODE"
 echo "=================================================="
+# Wrapper-emitted completion marker: pairs with PRELOOP_AGENT_EXEC_START.
+# The orchestrator treats this marker with exit code 0 as success without
+# requiring any model-printed sentinel.
+echo "PRELOOP_AGENT_EXEC_END:$GEMINI_EXIT_CODE"
 {post_exec_block}
 # Exit with gemini's exit code
 exit $GEMINI_EXIT_CODE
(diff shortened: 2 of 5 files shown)


d-mo:
Closing: wrong direction. The sentinel is a deliberate fail-closed contract added in 3902606e because agent CLIs exit 0 even when the agent died mid-task (see gemini.py continuation-check note). Trusting exit 0 reintroduces the false-success class. Follow-up will keep fail-closed semantics and add a second explicit positive-confirmation channel (result.json artifact from #231) instead.

PR #361, opened 2026-09-03
Flow runs: retry transient failures (runner 409, model disconnects) and record a failure category

d-mo:
Backend only. Evidence from all 100 FAILED runs on staging (2026-07-19 to 2026-09-02), classified from `error_message` plus each run's own timeline events. Full report in the private repo; the parts that matter:

## 1. Evidence

Pulled read-only from staging: `GET /api/v1/flows/executions?status=FAILED&limit=100` (100 rows, the API's maximum page and every failure since 2026-07-19), then `GET /api/v1/flows/executions/{id}/logs` for each. Classification is from `error_message` plus the run's own timeline events (milestones, `execution_retry`, `status_update`), not from the message alone.

| bucket | n | first seen | last seen | median duration | representative message | code path that raises it |
|---|---|---|---|---|---|---|
| `no_confirmation` | 33 | 2026-07-21 12:28 | 2026-08-24 13:12 | 0.8 m | `Agent exited with code 0 but did not produce the success sentinel — likely encountered an error.` | `FlowExecutionOrchestrator._monitor_agent_execution`, completion-contract fail-closed branch |
| `agent_error` | 17 | 2026-07-19 13:26 | 2026-08-11 09:14 | 5.7 m | `OpenCode command failed; see CLI output above.` / `OpenCode CLI exited with code: 1` | agent script exit code, read back by `ContainerAgentExecutor.get_result` |
| `model_transient` | 12 | 2026-08-15 10:00 | 2026-08-30 22:37 | 10.9 m | `message="stream error" providerID=preloop modelID=openrouter/deepseek/deepseek-v4-flash-0731 … "Upstream provider disconnected mid-stream: APIConnectionError: OpenrouterException - peer closed connection without sending complete message body"` | provider → gateway `_open_upstream_stream` → agent logs → `analyze_agent_failure` |
| `gateway_auth` | 10 | 2026-08-11 23:26 | 2026-09-02 23:51 | 4.3 m | `message="stream error" providerID=preloop … error.error="AI_APICallError: Invalid authentication credentials"` | **our own** `preloop/api/auth/jwt.py` rejecting the run's gateway token mid-run |
| `runner_conflict` | 8 | 2026-08-17 11:44 | 2026-09-02 20:07 | 6.2 m | `Failed to start agent Job: (409) Reason: Conflict … jobs.batch "agent-<execution_id>" already exists` | `ContainerAgentExecutor._start_kubernetes_pod` → `create_namespaced_job` (`container.py:652` on `main`) |
| `tool_error` | 8 | 2026-08-22 13:02 | 2026-08-24 10:08 | 9.0 m | `/bin/bash -lc 'pyspdxtools -i /workspace/sbom/tasmota-demo.spdx.json …' exited 1 in 258ms: ERROR:root:The document is invalid…` | a command the agent ran in its own workspace; surfaced by `_detect_error_in_logs` |
| `timeout` | 7 | 2026-08-03 13:31 | 2026-08-23 23:21 | 60.1 m | `Execution timed out after 3600 seconds` | `_monitor_agent_execution` wall-clock branch (`flow_orchestrator.py:3061` on `main`) |
| `model_config` | 4 | 2026-08-22 22:48 | 2026-08-24 03:55 | 0.3 m | `zai does not support parameters: ['parallel_tool_calls'], for model=glm-5.3` | litellm rejecting the request shape through the gateway |
| `runner_error` | 1 | 2026-08-22 12:36 | 2026-08-22 12:36 | 0.1 m | `exec /opt/entrypoint.sh: argument list too long` | pod start; oversized entrypoint script |

Flows most affected: `Pull Request Reviewer` (47), `RSA stock generic r3 verbatim` (17), `Merge Request Reviewer` (10).

**The raw table understates `model_transient` by 8.** Every one of the 8 `runner_conflict` rows started life as a transient failure whose 409 overwrote the original message (see §2.1).

---

### 2.1 `runner_conflict` (8) — self-inflicted, and it destroys the evidence of the real failure

Every one of the 8 has the same timeline. Execution `0a68540a-c7e7-4b5d-a0a0-5b31e401e4a7`:

```
20:13:48  agent_log_line     OpenCode CLI exited with code: 1
20:13:56  agent_status       FAILED
20:13:56  execution_retry    {"attempt":1,"max_attempts":2,"delay_seconds":15}
20:13:56  execution_warning  Attempt 1 of 2 failed with a transient upstream error; retrying.
                             (timestamp=… level=ERROR message="stream error" …)
20:14:11  status_update      FAILED  error_message="Failed to start agent Job: (409) Reason: Conflict"
```

The failure lands exactly `delay_seconds` after the retry was scheduled. Mechanism:

1. `_start_kubernetes_pod` derived the Job name from the execution id alone: `agent-{execution_id}`.
2. Agent Jobs are created with `ttl_seconds_after_finished = AGENT_JOB_TTL_SECONDS` (1 h), so attempt 1's Job still exists long after it exits.
3. `_run_agent_with_retries` starts attempt 2 with the same execution id → the same Job name → `409 AlreadyExists`.
4. The 409 propagates as `RuntimeError` out of `run()`, whose handler overwrites `error_message` with the 409 — the transient provider error that actually broke the run is lost.

So the retry mechanism that exists to rescue a transient failure is what converted it into a permanent one, and it deleted the evidence on the way out. Not "two dispatchers", not a stale Job from an unrelated run: a deterministic name shared by two sessions of the same execution.

**Second, latent instance of the same bug** (found by reading, not visible in the data because the affected runtimes were mostly Docker): the completion-confirmation nudge starts *another* agent session for the same execution from a copy of the same context. On Kubernetes that could only ever 409 while the first Job was inside its TTL.

### 2.2 `model_transient` (12) + the 8 above — providers drop streams

`stream error … peer closed connection without sending complete message body`, `Upstream provider disconnected mid-stream`, `exceeded retry limit, last status: 429`. The gateway already retried these (bounded, jittered, `Retry-After` honoured) when they happened before the first chunk reached the client. Two gaps: the retries were **invisible** (a rescued call looked identical to a slow one), and the budget was hardcoded, so an operator facing a bad provider hour had no lever.

### 2.3 Everything else

`no_confirmation` (33), `agent_error` (17), `tool_error` (8), `timeout` (7), `model_config` (4) are not transient infrastructure failures and are deliberately **not** retried — see §3.3. They were previously indistinguishable from each other without reading 100 messages by hand, which is the second problem this work fixes.

---

### 3.1 Kubernetes Job creation

- **Cause fixed:** each additional agent session of an execution now requests its own Job name — `agent-<id>-a2` for attempt 2, `agent-<id>-nudge` for the confirmation round — via `kubernetes_job_name(execution_id, session_suffix=…)`. Attempt 1 keeps the historic unsuffixed name so a run in flight across the deploy is still addressable by the session reference already stored on its row. Names are truncated to a valid DNS-1123 label with the suffix intact.
- **Bounded retry added:** `_create_kubernetes_job` retries `409`, `429` and `5xx` for `AGENT_JOB_CREATE_MAX_ATTEMPTS` (3) with exponential backoff plus jitter (jitter because the remaining conflict scenarios are caused by *concurrent* actors, and a fixed backoff keeps them in lockstep).
- **409 is resolved, not guessed at.** The conflicting Job is read: a live Job labelled `preloop.execution_id == this execution` is **adopted** (an idempotent create — a duplicate dispatch must not start two agents *or* fail the run); a finished leftover is deleted with `propagationPolicy=Background` (so its pods go with it), waited for, and the name reused; a Job owned by a **different** execution raises immediately instead of adopting or deleting someone else's agent.
- **Terminal errors stay terminal:** 403, invalid manifest, quota — raised at once, because retrying only delays a failure the user has to see.
- An execution is no longer marked FAILED by a 409 that resolves.

### 3.2 Model gateway

- Retry count is accumulated per request and consumed-and-cleared when the usage row is written (`meta_data.upstream_retries`), then published as **`retried: n`** on the `model_gateway_call` event, so a run rescued from a flaky provider is visible instead of merely slow.
- Attempts, backoff base and the `Retry-After` cap are now settings (`MODEL_GATEWAY_UPSTREAM_RETRY_*`), defaults unchanged (3 / 0.2 s / 8 s).
- Unchanged on purpose: a response that has already streamed bytes to the client is never retried.

### 3.3 "The agent process died" policy

Kept and documented rather than widened. A run is retried from scratch only when **all** hold: the agent exited non-zero (so the container's post-execution push / PR-creation block provably never ran), no actions were recorded on the timeline, and the executor's analysis of the *full* logs called the failure transient. An unknown exit code counts as unsafe. Everything else fails with a category. Re-running a review that already posted a comment is worse than not re-running it.

### 3.4 `failure_category`

New `flow_execution.failure_category` (String(32), nullable, indexed; migration `20260903_failure_category`), projected in the execution list and detail schemas, `openapi.yaml` regenerated. Vocabulary: `runner_conflict`, `runner_error`, `model_transient`, `model_auth`, `model_quota`, `model_config`, `no_confirmation`, `tool_error`, `agent_error`, `timeout`, `cancelled`, `unknown`.

Derived once at failure time, in order of authority: explicit category from the raising code (`AgentStartError`) → structural message shapes that identify the failing layer regardless of provider noise (runner conflict, runner error, timeout, cancellation, missing confirmation) → the executor's full-log analysis verdict (sharing the `upstream_errors` taxonomy with the gateway) → provider message patterns → `unknown`. Unmatched failures stay `unknown` rather than being folded into `agent_error`, so the gap is measurable. Not backfilled: categories come from live failure context, and inferring them from historical prose would launder a heuristic into stored data.

### 3.5 Configuration & docs

`AGENT_JOB_CREATE_MAX_ATTEMPTS` (3), `AGENT_JOB_CREATE_RETRY_BASE_SECONDS` (0.5), `MODEL_GATEWAY_UPSTREAM_RETRY_MAX_ATTEMPTS` (3), `MODEL_GATEWAY_UPSTREAM_RETRY_BASE_SECONDS` (0.2), `MODEL_GATEWAY_UPSTREAM_RETRY_AFTER_CAP_SECONDS` (8) — settings + env + helm values, wired into the api, flow-execution worker and gateway deployments, defaults on. Either `maxAttempts: 1` disables that layer. New "Transient Failure Handling" section in `docs/architecture/flows.md`.

### 3.6 Tests

21 Job-creation tests (mocked k8s client) covering adopt / delete-and-recreate / vanished / foreign-owner / 429 / 5xx / 403 / attempt budget / DNS-1123 validity; a regression test that a retried run produces two *different* Job names (on `main` both are `agent-<id>`); the nudge's own session name; 35 category-derivation tests keyed to the real staging messages above; end-to-end category storage through `orchestrator.run()`; gateway retry-count propagation into `meta_data` and the event. `3898 passed` over `backend/tests/{agents,services,models}` plus `test_flow_orchestrator.py`; the migration applies cleanly from base on a fresh database.

---

## 4. Recommended, not implemented

1. **`no_confirmation` is the biggest bucket (33/100) and is mostly a false negative.** These runs exited 0 after a median of 0.8 minutes and simply never printed the sentinel; two flows account for 30 of them. That is Preloop's completion contract failing, not the agent — the work may well have been done. Now that it is a distinct, countable category, the fix is a prompt/preset change (and possibly accepting a wider set of completion signals) rather than more retries. Retrying it would be actively wrong: exit 0 means the container's push/PR block already ran.
2. **`gateway_auth` (10) is ours, not a provider's.** The message is our own `preloop/api/auth/jwt.py` `Invalid authentication credentials`, returned to an agent whose `providerID=preloop`. Failures land 1.5–30 minutes into runs, while the runtime token is minted with a 2-hour expiry — so this is not plain expiry. Candidates: `_cleanup_temporary_api_token` firing while a session is still live, or a token invalidated by a concurrent run of the same flow. Deserves its own investigation; a retry here would be useless anyway (`model_auth` is deliberately non-retryable).
3. **`timeout` (7) all sit exactly at the 3600 s ceiling** with a 60.1-minute median — the ceiling *is* the cause. A per-flow timeout budget (and a warning as the budget is approached) would separate "genuinely long" from "stuck".
4. **Surface `failure_category` in the console** (the API now returns it) and alert on it: `runner_conflict`/`runner_error` should be ~0 after this change, and a rising `unknown` share is the signal to extend the vocabulary.
5. **Delete the agent Job on terminal status** instead of relying on the 1-hour TTL. The TTL is what made the name collision possible; a positive cleanup would also free cluster resources sooner. Not done here because the TTL is also what lets recovery and log rescans read a finished pod.
6. **Record the per-attempt error history on the execution**, not just the last one. Today a retried run keeps only the final message; the earlier attempts survive only in the timeline.

---

## Appendix: method

Authenticated to staging with the credentials at `/Users/dimo/git/spacecode/preloop-ee/tmp/staging-creds.json` via a Playwright login at `/login`, then issued `fetch` calls from the logged-in page using the JWT from `localStorage`. Part 1 was strictly read-only: two `GET` endpoints, no writes. Artifacts (raw execution list, per-execution timeline events, classification script) are in `/tmp/flowev/`.

> [!NOTE]
> **Medium**
> Backend-only change to retry transient flow failures (K8s Job 409s, provider stream drops) and record a machine-readable `failure_category`; well-tested but touches critical execution/gateway retry logic.
>
> **Overview**
> Fixes the self-inflicted `runner_conflict` (deterministic Job-name collision on retry) via per-session Job names and bounded jittered create retries with adopt/delete conflict resolution, makes gateway upstream retries configurable and visible (`retried`), and adds an indexed `flow_execution.failure_category` column with a closed vocabulary derived at failure time.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit e48ed7e. Updates automatically on new commits.</sup>


d-mo:
## Review fixes pushed

All five review findings are addressed (two raised here, three on the GitLab mirror of this branch, which is now closed in favour of this PR). Branch head: `3af14f39e3c6bcc5ec1a9c9504f36cbf03249e17` (`3af14f3` only drops em dashes from the prose this branch adds; no behaviour change).

| # | Finding | Where | Fix | Commit |
|---|---|---|---|---|
| 1 | `AGENT_JOB_CREATE_MAX_ATTEMPTS=1` still deleted a finished leftover Job on 409 and then failed without recreating | `backend/preloop/agents/container.py:795` | The delete is a precondition for another create, so it is gated on one remaining (`may_delete`). At the attempt budget's end the leftover and its logs stay and the run fails as `runner_conflict`. Adoption of a live Job is not gated: it needs no further attempt. | `089706d` |
| 2 | Ownership guard silently adopted or deleted a conflicting Job with no `preloop.execution_id` label | `backend/preloop/agents/container.py:869` | Unknown ownership now fails exactly like foreign ownership, so the guard is fail-closed. Every Preloop creation path sets the label, and did so before this branch, so no in-flight Job of ours lacks it. | `089706d` |
| 3 | `AgentStartError.retryable` set but never read | `backend/preloop/agents/errors.py:39` | Dropped. The only bounded retry of a Job start lives inside the runner, and the flow-level retry is scoped to the "agent process died" policy, so a flag nothing acts on only misleads. The docstring now says that explicitly. | `e816bad` |
| 4 | `_last_upstream_retry_count` could leak to the next request when `_record_gateway_request` was not reached | `backend/preloop/services/openai_gateway.py:664` | All six request entry points (chat completions, responses, Anthropic messages, and the three streaming variants) re-arm the counter through `_begin_request_accounting`, so it is request-scoped no matter how the previous request ended. Telemetry only, nothing about what is retried changed. | `37fcc78` |
| 5 | Module docstring stated the wrong derivation precedence; `_MESSAGE_RULES` unused | `backend/preloop/services/flow_failure_category.py:11` (docstring), `:240` (unused global, from the code-quality bot) | Docstring now matches the code: explicit category, then structural message shapes, then the executor's full-log analysis verdict, then provider-message patterns. `_MESSAGE_RULES` removed; the derivation walks the structural and provider tuples separately. | `6c5c2ed` |

### New tests

`backend/tests/agents/test_container_job_create.py`
- `test_single_attempt_setting_keeps_the_leftover_on_conflict`: 409 plus a finished leftover at `maxAttempts=1` raises `runner_conflict`, deletes nothing, creates once.
- `test_single_attempt_setting_still_adopts_a_live_job`: the disabled retry still adopts a live Job.
- `test_last_attempt_never_deletes_the_leftover`: with 3 attempts, exactly 2 deletes, never on the final one.
- `test_conflict_with_unlabelled_live_job_fails_closed` and `test_conflict_with_unlabelled_finished_job_is_not_deleted`.

`backend/tests/services/test_openai_gateway_retry.py`
- `test_stale_retry_count_is_cleared_when_a_request_starts` and the streaming variant: a count left behind by a request that never recorded is cleared by the next request.
- `test_every_request_entry_point_rearms_the_retry_count`: a new entry point that forgets the reset re-opens the leak.

### Verification

```
pytest backend/tests/agents backend/tests/services backend/tests/models backend/tests/test_flow_orchestrator.py
3906 passed, 5 skipped in 122.67s
pytest backend/tests/endpoints/test_openai_gateway.py test_anthropic_gateway.py test_gemini_gateway.py test_flows.py test_gateway_streaming_accuracy.py
103 passed
```

`pre-commit run --files <changed>` passes; `openapi.yaml` regenerates unchanged. `docs/architecture/flows.md` layer 1 documents the delete gate, the missing-label case and what `maxAttempts=1` now means.

d-mo:
Verified this branch's `failure_category` vocabulary against the 10 `gateway_auth` staging failures investigated in #363 (2026-08-11 23:26 to 2026-09-02 23:51). No adjustment is needed: all 10 derive as `model_auth`.

Two distinct message shapes appear in that bucket, and both land correctly:

```
timestamp=... level=ERROR message="stream error" providerID=preloop modelID=deepseek/deepseek-v4-pro
  ... error.error="AI_APICallError: Invalid authentication credentials"          -> model_auth  (7 runs)
ERROR: Reconnecting... 5/5
ERROR: unexpected status 401 Unauthorized: Invalid authentication credentials    -> model_auth  (3 runs)
```

The second shape is the interesting one: it matches `_MODEL_TRANSIENT_RE` too (`Reconnecting...`, and the first shape matches `stream error`), and it only comes out as `model_auth` because `_MODEL_AUTH_RE` precedes `_MODEL_TRANSIENT_RE` in `_PROVIDER_MESSAGE_RULES`. That ordering is load-bearing, since `model_auth` is deliberately non-retryable and `model_transient` is not. Worth keeping the two staging shapes above as fixtures if this file grows.

Method: replayed each execution's stored `error_message` plus its full agent log through `analyze_agent_failure` and `derive_failure_category(status="FAILED", ...)` on this branch. The log analysis returns no verdict for these runs (`error_class=None`, `transient=False`), so the provider-message rules are what classify them.

Cause of those 10 failures, for context: a deploy drain cancelled the worker task monitoring the run, and the orchestrator's `finally` revoked the runtime API key of an agent that was still running, so our own auth layer rejected its next model call. Fixed in #363.

PR #366, opened 2026-09-03
Console UX wave 9: failure category chips, progressive Users/Agents loading, Usage keeps numbers while updating, price override counts as priced with a reprice offer

d-mo:
Console UX wave 9, branched from wave 8 before #360 merged. Spec: `product/ux-review/2026-09-03/CONSOLE-WAVE-9-SPEC.md` (private repo). Frontend only, 1313 tests on the branch.

- **Failure category** (from #361's `failure_category`) as a soft chip after the status pill on the executions list, run page strip, attention evidence, Inventory Flows tab, and feed; grouped summaries ("5 failed: 3 model transient, 2 no confirmation"); absent field renders nothing.
- **Progressive loading**: Users and Agents tabs render identities first and fill usage cells as the summary arrives; Spend sort waits for usage.
- **Usage card keeps its numbers** while a new range loads (60% opacity, header spinner, "from rollup" label, a "Long ranges take longer" hint for 1y after 2s).
- **A price override is a price, including $0**: attention no longer calls an overridden model unpriced; saving a price offers "Apply to past usage since <effective date>" via the existing reprice endpoint, with plain copy that recorded usage keeps its cost until repriced. This is the ox-alpha case.

**Merge note:** `frontend/src/views/authed/dashboard-control-plane-view.ts` and its test conflict with the #360 review fixes now on `main` (Users tab permission gating vs progressive loading). Both sides are small; resolve by keeping the permission gate and the progressive-loading render together. Not resolved here because the session was being wrapped up.

> [!NOTE]
> **Low**
> Frontend-only UX wave 9; both prior review findings (permission gate, reprice window) are fixed.
>
> **Overview**
> Failure-category chips across five surfaces, progressive Users/Agents loading, a dim-and-keep Usage card, and price overrides treated as real prices (including $0) with a reprice backfill offer. The #360 `view_users` gate is now preserved alongside progressive loading, and a future `effective_from` no longer offers an empty reprice window.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 6cee62c9. Updates automatically on new commits.</sup>


d-mo:
Merged `main` (`155d27df`) and addressed the review comments in `6cee62c9`:

1. **Users tab `view_users` gate** — wave 9's progressive Users fetch is kept, and the request/tab still require `view_users` (from #360). A reader without the permission does not get a Users tab or a `GET /users` 403.
2. **Future `effective_from`** — saving a price that starts in the future no longer offers "Apply to past usage" (that window would have been empty).

PR #397, opened 2026-09-04
Tools: accept agent tool sources in policy documents and keep refresh as a scan

d-mo:
## Why
On the Tools page, "Suggest starter policy" returned `Internal server error: 1 validation error for PolicyDocument ... Tool 'Bash' references unknown MCP server 'agent'`, and the refresh button on an MCP server card opened that same failing dialog.

## Root cause
- `PolicyDocument.validate_references` only knew `builtin`, `mcp`, `http`. Native agent tools are stored as `ToolConfiguration.tool_source="agent"`, and `export_current_policy` copies that into `ToolDefinition.source`, so the validator treated `agent` as an MCP server name.
- `_handleScanMCPServer` queued a starter policy suggestion before the scan, so a plain refresh opened the dialog.

## What changed
- `ToolSource.AGENT = "agent"` and `is_known_tool_source()`; the validator accepts every `ToolSource` value case-insensitively (schema.py). Same allowlist in `loader.py` and `policies.py`; applying a document with `source: agent` stores `tool_source="agent"` instead of looking up a server named agent.
- A starter policy for MCP server X now drops agent-sourced tools and other servers' tools from the prompt context and the generated document (`policy_generation.py`). General edits still restore unrelated tools.
- Frontend: refresh only scans and reloads; it never sets `_pendingStarterPolicyServerId`.

## Tests
- `tests/services/test_policy_schema.py` + `test_policy_generation.py`: 79 passed (source `agent` and `AGENT`; starter policy with a native Bash tool and another server's tool present produces only the target server's tools, no ValueError).
- Frontend: 1412 passed, 0 failed (refresh dispatches `scan-server` only, POSTs `/mcp-servers/{id}/scan`, does not open the dialog or call `/policies/generate`).
- `openapi.yaml` unchanged.

> [!NOTE]
> **Low risk**
> Well-tested, minor changes to policy validation and generation with no security implications.
>
> **Overview**
> Teaches the policy validator/loader to accept the `agent` tool source, scopes starter-policy generation to the target MCP server, and keeps the refresh button as a scan-only action.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit dfbab102. Updates automatically on new commits.</sup>

### DIFF
--- a/backend/preloop/api/endpoints/policies.py
+++ b/backend/preloop/api/endpoints/policies.py
@@ -1155,6 +1155,16 @@ class GeneratePolicyRequest(BaseModel):
             "for the LLM (recommended for more accurate generation)"
         ),
     )
+    scope_mcp_server_name: Optional[str] = Field(
+        None,
+        description=(
+            "When set, starter-policy generation scopes LLM context and "
+            "raw model output to this MCP server. The merge still restores "
+            "all current tools so the diff preview only shows this "
+            "server's changes. Older clients that omit the field fall "
+            "back to matching the prompt text."
+        ),
+    )
 
 
 class GeneratePolicyFromAuditRequest(BaseModel):
@@ -1232,7 +1242,10 @@ async def generate_policy(
         # are not thread-safe and must not be shared across threads.
         model = service._resolve_model()
         context_block = (
-            service._build_context_block(request.prompt)
+            service._build_context_block(
+                request.prompt,
+                scope_mcp_server_name=request.scope_mcp_server_name,
+            )
             if request.include_current_config
             else ""
         )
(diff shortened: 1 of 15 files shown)


d-mo:
Review follow-up (d53b05f9, 2c5926c1): the starter-mode merge now restores every current tool again, so `/policies/diff` reports only the target server's changes instead of N removals; the Tools view sends an explicit `scope_mcp_server_name` on `GeneratePolicyRequest` (the prompt regex stays as a fallback for older clients; openapi.yaml regenerated); context-block scoping has its own test; editor icon buttons got accessible labels. Backend 113 passed, frontend 1412 passed.

PR #406, opened 2026-09-05
Overview: paint the Inventory lists first, fill usage as it arrives, never flash Next steps

d-mo:
## Why
On staging the Inventory box waited for every request in the fold before drawing a row, and the "Next steps" checklist could appear while the lists were still loading because empty arrays before the first response read as "the account has nothing".

## What changed (`dashboard-control-plane-view.ts`, `inventory-card.ts`)
- Each Inventory tab paints as soon as its cheap list request lands (`GET /agents`, `/flows`, `/ai-models/overview`, `/tools`, `/users`), with the usage and cost columns as skeleton cells. Usage fills in when the summary breakdown and executions arrive, in arrival order (D32), without re-rendering rows back to skeleton.
- Next steps renders only once the list phase has resolved and reports zero rows for the relevant kind; while lists load it renders nothing. A finished account stays finished (D31).
- The per-tab sort select is disabled until usage arrives, so it cannot sort on columns that are still skeletons; the pending Runs cell is a placeholder, not a nameless link.
- Request count on first paint unchanged; one duplicate `GET /ai-models` removed (the page and a child component both asked). Socket refresh path unchanged from #381.
- Dropped the write-only `fetchingInventory` flag; list loading flags are cleared in their own `finally`.

## Tests
`dashboard-view.test.ts`: rows render before usage arrives; usage fills in without unmounting rows; Next steps hidden while loading and shown only on a resolved empty inventory; request count assertion measured against the same URL filter the report uses. Full suite: 130 files, 1418 passed, 0 failed.

Founder request 2026-09-05: "render the cheap stuff first, then calculate and show their usage in discrete API calls". Analysis of the request set is in the implementer report (section 1 table).

> [!NOTE]
> **Low Risk**
> Progressive rendering refactor with no security surface, behaviour-preserving request reorganisation, and thorough test coverage; the two minor cleanup findings are now resolved.
>
> **Overview**
> Splits the Overview Inventory into a fast identity-list paint followed by usage/cost cells filling in arrival order, and keeps the "Next steps" checklist hidden until the list phase resolves. Verified the duplicate `/ai-models` removal and the fetch reorganisation against the pre-PR source; the write-only next-steps flag and the flow usage coupling were both removed in the latest commit.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 627efe39. Updates automatically on new commits.</sup>

### DIFF
--- a/frontend/src/components/inventory-card.test.ts
+++ b/frontend/src/components/inventory-card.test.ts
@@ -91,6 +91,7 @@ async function card(
     loading: boolean;
     usageLoading: boolean;
     usageLoadingFlows: boolean;
+    usageLoadingFlowCost: boolean;
     usageLoadingModels: boolean;
     usageLoadingTools: boolean;
     flowRunsCapped: boolean;
@@ -113,6 +114,7 @@ async function card(
       .loading=${props.loading ?? false}
       .usageLoading=${props.usageLoading ?? false}
       .usageLoadingFlows=${props.usageLoadingFlows ?? null}
+      .usageLoadingFlowCost=${props.usageLoadingFlowCost ?? null}
       .usageLoadingModels=${props.usageLoadingModels ?? null}
       .usageLoadingTools=${props.usageLoadingTools ?? null}
       rangeLabel="30d"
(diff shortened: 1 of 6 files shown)


d-mo:
Addressed the two low findings in 627efe39:

1. **Write-only next-steps remember.** Removed `NEXT_STEPS_DONE_KEY`, `readCoreStepsDone`, and `rememberCoreStepsDone`.
2. **Runs/Failed waiting on `$`.** `usageLoadingFlowCost` is separate from the executions pending flag, so Runs and Failed fill when executions land.

Tests: `inventory-card.test.ts` + `dashboard-view.test.ts` (84).

PR #416, opened 2026-09-05
Read runner-backed execution logs from the database and halt runners on stop

d-mo:
## Problem

The flow execution logs endpoint treats every running execution with an `agent_session_reference` as container backed. Executions leased to a private runner carry a lease reference (`runner:{runner_id}:{execution_id}`, or `runner:queued:{pool}:{execution_id}` while waiting), not a container id or Kubernetes Job name. The pod lookup then builds an invalid label selector (`job-name=runner:...`: colons, over 63 characters), the Kubernetes API returns 400, the error is swallowed, and the endpoint answers `source: container` with zero lines. The runner's output, which streams over the runner WebSocket into `flow_execution_log`, stays invisible in the console until the execution finishes, and every poll of the log pane generates API error noise. Queued executions are hit twice: they show nothing at all, hiding the queue-status entries that would explain the wait.

The stop command had the same container assumption with two extra effects:

- the final-log fetch 400'd the same way (harmless but noisy), and
- the halt was never flagged on the runner (`RemoteRunnerExecutor.stop` sets `halt_requested`, but the endpoint built a `ContainerAgentExecutor` directly), so cancelling a runner-backed execution left the job running on the runner until it finished or the runner process was killed.

## Changes (`api/endpoints/flows.py`)

- Logs endpoint: skip the container path when the session reference parses as a runner reference (`runner_id_from_session_reference` / `pool_from_session_reference`, both already imported) and serve the database logs instead.
- Stop command: for leased executions, flag `runner.halt_requested` (the runner learns it via its heartbeat ack) instead of constructing a container executor; for queued executions, skip both paths since nothing runs yet and the status update alone releases the orchestrator's queue wait.

## Tests (measured)

- 4 new endpoint tests: runner-leased and queued executions read logs from the database without constructing an agent executor; stop on a leased execution flags the halt and builds no container executor; stop on a queued execution touches neither.
- `backend/tests/endpoints/test_flows.py` plus the runner-related endpoint, agent, and service modules: 147 passed / 0 failed.
- pre-commit clean (ruff-format reformatted one test).

> [!NOTE]
> **Low**
> Bug fix with no security impact and thorough test coverage; all findings addressed.
>
> **Overview**
> Routes runner-backed and queued flow executions around the container log path (serving DB logs) and flags `runner.halt_requested` on stop instead of building a container executor. Reuses existing session-reference helpers and mirrors `RemoteRunnerExecutor.stop` semantics.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 2685e5f. Updates automatically on new commits.</sup>

### DIFF
--- a/backend/preloop/api/endpoints/flows.py
+++ b/backend/preloop/api/endpoints/flows.py
@@ -769,9 +769,13 @@ async def get_flow_execution_logs(
     skip: int = 0,
     limit: int | None = None,
 ) -> Dict[str, Any]:
-    """Get execution logs from the container (if running) or database (if finished).
+    """Get execution logs from the live container or from persisted database rows.
 
-    For running executions, fetches logs directly from the Docker/Kubernetes container.
+    For running container-backed executions, fetches logs directly from the
+    Docker/Kubernetes container. Runner-backed executions (lease references
+    such as ``runner:{runner_id}:{execution_id}``) always return persisted
+    database logs, even while they are still running, because their output
+    arrives over the runner WebSocket rather than a container log stream.
     For finished executions, returns persisted logs from the database.
 
     Args:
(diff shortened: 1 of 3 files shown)


d-mo:
Addressed the docstring finding: `get_flow_execution_logs` now states that runner-backed (and queued) executions read persisted database logs even while they are still running. Container-backed running executions still come from the live container stream.

PR #492, opened 2026-09-07
Console polish: agents page geometry, one token total in the lists, bell at the header's optical size

d-mo:
Three console polish items from the 2026-09-07 founder review, all in the frontend.

## 1. The agents page sits in the same box as the rest of the console

`agents-view` carried `.console-page` (the shared page box: 80rem, centred, 2rem
side inset) on its header band, its list wrapper and its cards grid, in every
mode. That is right only on the canvas, which asks the shell for the whole window
(`request-full-bleed`) and so has no shell padding to inherit. In the default list
the shell was already centring the column and paying the inset, so the page paid
it twice.

Measured in a browser (web-test-runner, real Shoelace tokens, `getBoundingClientRect`),
inset from the view's own left edge to its title:

| viewport column | agents before | agents after | flows | tools | approvals |
| --- | --- | --- | --- | --- | --- |
| 1440 | 112px | 0px | 0px | 0px | 0px |
| 1280 | 32px | 0px | 0px | 0px | 0px |

Inside the shell the doubled part is the 2rem, so the agents list card also came
out 64px narrower than the Flows one. The fix removes the divergence rather than
adding an override: a `pageBoxClass` getter applies `.console-page` only on the
canvas. Agents joins the shared `page-width.test.ts`, and a new
`agents-page-geometry.test.ts` pins the agents header and list card to the Flows
ones within 1px at 1440 and 1280, and pins the canvas keeping its own box.

## 2. The default lists state one token total

The lists carried the full breakdown ("12.4K in / 3.1K out / cache 68% hit"). It
does not fit a list column: at the agents table's 190px tokens column a long
figure lost the end of the cache segment to the cell's ellipsis.

`token-figures` gets a `total-only` mode: one compact figure, with the same
breakdown sentence still in the tooltip and in `title`. Detail views, cost tables,
the usage card and the session panel keep the breakdown on the surface. Sorting is
by the total, which is what the cell now states.

Widths that follow: agents tokens column 190px to 110px, flows 170px to 96px, and
the agents table `min-width` 1340px to 1260px on the same PR #479 arithmetic
(pixel columns now sum to 720px, owner + model still 24%, so the auto Agent column
is `0.76 x width - 720` = 238px at 1260, above the 212px it needs). The #479 test
passes unchanged.

Configurable columns and composite columns with per-part sorting are not in here.

## 3. The notification bell is at the optical size of the header icons

The header row is the nav toggle (a 24px glyph), the bell and a 32px avatar. The
bell was `font-size: 1.8rem`: a 28.8px glyph in a 45px box, the loudest thing in a
quiet header.

Candidates (bell, bell-fill, inbox, app-indicator, at 1.25 / 1.375 / 1.5rem) were
rendered side by side with real Shoelace icons and measured. The glyph stays the
plain `bell` (bell-fill reads as a ringing state and competes with the red badge,
inbox reads as mail, app-indicator is illegible at these sizes and duplicates the
badge's dot); only the size changes, and it changes on the icon button rather than
becoming a one-off icon. 1.375rem is 22px of glyph in a 38px box beside a 32px
avatar. A bell fills its Bootstrap Icons box where a hamburger's three short rules
do not, so optical parity is below the nav toggle's size, not equal to it.

Badge placement survives: it is still positioned on `.notification-button`, and a
test pins it to the button's top-right corner.

## Tests

- Full frontend suite: 169 files, 2083 passed, 0 failed.
- New/changed: `agents-page-geometry.test.ts` (3), `page-width.test.ts` (+agents),
  `agents-view.test.ts` 43 passed, `flows-view.test.ts` 49 passed,
  `token-figures.test.ts` 18 passed, `console-header.test.ts` 35 passed.
- `npm run build` passes. Bundle 3,083.88 kB to 3,085.57 kB (gzip 699.71 to 700.44 kB).
- `npx tsc --noEmit`: 251 pre-existing errors on the base commit, 251 on this
  branch (no change; none in the files touched here).

> [!NOTE]
> **[Low]**
> Frontend-only console polish (agents page geometry, one token total in the lists, bell sizing); no security or API surface changes and well covered by browser-measured tests.
>
> **Overview**
> Three founder-review polish items: agents page geometry is brought in line with the shared shell box (removing a doubled `.console-page` inset), the default agent/flow lists state a single token total (breakdown moves to tooltip/`title`), and the notification bell is resized to optical parity with its header neighbours. The token-column sort now uses the same `totalTokensOf` helper the cell renders, resolving the one prior MEDIUM suggestion. All changes are frontend, with new geometry and total-only tests.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit dfc67157. Updates automatically on new commits.</sup>

### DIFF
--- a/frontend/src/components/console-header.test.ts
+++ b/frontend/src/components/console-header.test.ts
@@ -15,6 +15,7 @@ import {
 import './console-header.ts';
 import type { ConsoleHeader } from './console-header.ts';
 import { publishAttentionSummary } from '../utils/attention-summary';
+import { loadShoelaceTokens } from '../utils/test-shoelace-theme';
 import { Router } from '@vaadin/router';
 
 const USER = {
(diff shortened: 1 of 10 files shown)


d-mo:
Addressed the token-column sort: agents and flows now sort with `totalTokensOf`, the same helper the cell uses, so a row with in/out counts and no `total_tokens` no longer sorts as zero. Tests cover that split-payload case.

PR #495, opened 2026-09-07
Multi-select without the jump: the bulk bar takes over the list toolbar

d-mo:
## The complaint

Prod, 2026-09-07: "The multiselection lists behave weirdly. When I select an
item the entire table is pushed down because the action bar appears."

True on every list with multi-select. The bar was rendered as a sibling above
the table and returned `nothing` at zero selected, so ticking a box inserted a
row into the page. Measured on the agents list (table `getBoundingClientRect().top`):
150px idle, 168px with one row picked, 174px after select all. That is an 18px
jump under the pointer at the moment of the click.

## What this does

The bar stops asking for a row of its own and takes over a row the page already
has.

- **New `list-bar-swap`**: both layers sit in one CSS grid cell, so the row's
  height is `max(base, bulk)` and cannot depend on the selection. The layer that
  is not in front gets `visibility: hidden` (keeps its geometry, leaves the a11y
  tree) plus `inert` (leaves the tab order). Nothing is unmounted, so search text
  and filters come back exactly as they were.
- **`list-toolbar` gained a `bulk` slot** and a `selecting` flag. Agents, flows
  and AI models put their bar there: search, filters and the view switcher step
  aside, the table header stays, so sorting still works while a selection is up
  (the Mist portal pattern).
- **Fallback for a page with no toolbar** (API keys): the bar is absolutely
  positioned over the `thead`, which leaves the header row height untouched
  (the Gmail pattern).
- **Approvals** has neither a toolbar nor a table header, so its "Waiting for
  you" heading plays the same part.
- Escape clears the selection and the toolbar returns. The bar keeps
  `role="toolbar"` and now has exactly one live region reading "3 selected" and,
  during a run, "2 of 7".
- Action buttons render through `resource-actions`, the shared renderer, so a
  bulk action and a row action of the same id look and collapse the same way.
  That change is additive (`size`, `data-action`); `actions`, `actionsSets` and
  the intersection logic are untouched, so the action-registry branch merges
  cleanly on top.

Cards mode uses the same toolbar swap, so it does not move either.

## Numbers

| agents list, table top | idle | one selected | select all | cleared |
| --- | --- | --- | --- | --- |
| before | 150px | 168px | 174px | 150px |
| after | 150px | 150px | 150px | 150px |

Real browser tests assert an unchanged `top` for the agents table, the agents
cards grid, the flows table, the approvals request list and the api-keys table
body, on select, on select all and on clear, plus that clearing gives the search
text and the active filters back.

- Whole frontend suite: 169 files, 2087 passed, 0 failed, 20.5s (main: 168
  files, 2055 passed). 32 new tests.
- `npx tsc --noEmit`: 108 errors on this branch, the same 108 on main, none in
  the files touched here.
- `npm run build` passes. `main-*.js` 3,083.88 kB to 3,087.98 kB raw, 699.71 kB
  to 700.70 kB gzipped: +0.99 kB gzip.
- Motion: one `opacity 120ms ease-out`, no geometry animated, behind a
  `prefers-reduced-motion: reduce` guard. Inside the DESIGN.md console budget.

## Alternatives considered

- **Floating sticky bar at the bottom of the viewport.** No shift, but it covers
  the last rows and fights the mobile bottom nav.
- **Always reserving the bar's row.** No shift either, but every list pays a
  permanently empty row for a state that is rare.

Happy to revisit either if the toolbar takeover reads as too magical.

> [!NOTE]
> **Low**
> Frontend-only UI refactor with no security or correctness risk, backed by 32 new browser tests asserting an unchanged layout on select/select-all/clear across every affected list.
>
> **Overview**
> Eliminates the multi-select layout jump by docking the bulk action bar into each list's existing toolbar row via a new `list-bar-swap` grid overlay, rather than inserting a new row on first selection. Adds a "Select all N" affordance and renders bulk actions through the shared `resource-actions` renderer; search/filter state survives the swap and Escape restores it.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 6638837c. Updates automatically on new commits.</sup>


d-mo:
Addressed the remaining low notes:

- **Dead CSS `.bulk-bar-slot`**: removed the orphaned rule from `frontend/src/styles/console-styles.css`. Grep across `frontend/` shows zero remaining HTML/class usages.
- **CHANGELOG**: added an Unreleased Fixed bullet for the no-shift toolbar swap and the bulk bar's "Select all N" affordance (this PR had no CHANGELOG entry yet).

PR #504, opened 2026-09-08
Signed outbound event webhooks: endpoints, outbox delivery and seven v1 events

d-mo:
# Signed outbound event webhooks

## Why

There is exactly one outbound webhook in the product today: `approval_service.py`
reads `webhook_url` from `approval_workflow.approval_config` and posts it once,
unsigned, on the request path, with a 10 s timeout and no retry. Nothing fires
when an approval is decided, when a policy denies a call, when a session ends,
when a budget is crossed or when a flow execution finishes. Two partner
conversations in one week (a GRC platform and a legal validation firm) hit that
gap, and any SIEM or GRC integration needs the same thing.

## What this adds

**Account-level webhook endpoints.** CRUD at `/api/v1/event-webhooks`, in OSS
and EE, gated on `view_policies` / `manage_policies`. The signing secret is
returned exactly once on create and is never readable again; reads carry a
four-character hint.

**Seven v1 events**, each emitted from a chokepoint that already existed rather
than from new instrumentation:

| Event | Emitted from |
| --- | --- |
| `approval.created`, `approval.decided` | `ApprovalService._broadcast_approval_update` (covers approved, declined, expired, cancelled, with actor kind and the rule that decided it) |
| `policy.denied` | `PolicyEvaluator._log_policy_decision_async` |
| `session.ended` | the three places a runtime session is closed (execution finished, idle sweep, operator) |
| `budget.threshold`, `budget.exceeded` | the gateway's existing budget snapshot |
| `flow.execution.finished` | `FlowOrchestrator._notify_terminal`, carrying the evidence receipt from #483 when one exists |

Stable envelope: `id`, `type`, `version`, `occurred_at`, `account_id`, `data`.
Event ids are deterministic (UUIDv5 over a natural key), so emitting the same
fact twice produces one delivery and receivers can deduplicate on
`X-Preloop-Event-Id`.

**Signature.** HMAC-SHA256 over `timestamp + "." + body`, sent as
`X-Preloop-Signature: t=...,v1=...` with a fresh timestamp on every attempt and
a documented 5-minute tolerance.

**Delivery.** A transactional outbox written inside the caller's transaction and
drained by a worker, so nothing HTTP happens on the request path. At-least-once,
6 attempts over about an hour (10 s, 1 m, 5 m, 15 m, 40 m, jittered 0.8 to 1.2),
dead-letter state visible in the API and the console, replay of one event id as a
new generation that leaves the original rows on the record, a per-endpoint
circuit breaker after 10 consecutive failures, and a bounded queue (10,000
undelivered rows per account, 64 KB per payload, 14-day retention).

**Console.** `Settings > Webhooks`: the endpoint list with its filter and state,
create with the event catalogue fetched from the server, test send, and recent
deliveries with a Replay on the dead ones.

**Docs.** `docs/guide/webhooks.md`: the envelope, verification samples in Python
and Node that sign the raw body, retry semantics, and a plain list of what the
signature does **not** prove (the audit log is not signed, the HMAC is symmetric
so it is not non-repudiable, and it adds no assurance to the evidence receipt it
carries).

## Behaviour changes

The existing approval webhook is migrated onto the new path with a compatibility
shim. `approval_config.webhook_url` keeps working and the **body is unchanged**,
posted verbatim rather than wrapped in a v1 envelope. Two things do change:

1. It is now **signed and retried**. The secret is
   `approval_config.webhook_secret`: set it to choose it, or leave it out and
   Preloop generates one and writes it back beside the URL. A receiver that
   ignores the signature header keeps working; a receiver that was relying on
   exactly one POST will now see retries after a non-2xx.
2. `approval_request.webhook_posted_at` is now stamped when a receiver
   **accepted** the delivery, not when the POST was issued. Anything reading that
   column as "we tried" now reads it as "it arrived".

Shim endpoints appear in the console list read-only, so an operator can see one
failing, and are edited on the approval workflow.

## Also fixed

`webhook_block_private_targets` was defined in config and read by nothing. The
delivery worker runs inside the cluster, which makes an operator-supplied URL a
request forgery primitive. It now resolves the host at registration and refuses
loopback, link-local, private and reserved space. Still off by default, because
a self-hosted deployment posting to a collector on its own private network is the
normal case; the docs state that it resolves once and so does not catch a name
that later re-resolves into private space.

## Tests

134 new backend tests, all passing: signing (12), outbox routing, retry schedule,
circuit breaker, bounded queue and replay (33), one envelope-shape test per event
type (16), the worker and what actually goes on the wire (26), the shim (10), the
private-target guard (12), the API surface (21), and a migration downgrade/upgrade
cycle against real Postgres with the rebuilt columns diffed against the ORM (4).
`test_approval_service.py` was rewritten around the new path and holds at 119.

- Backend `tests/services tests/api tests/models`: 4,951 passed, 13 failed.
  All 13 reproduce unchanged on `main` in a clean worktree (log-assertion and
  gateway-payload tests unrelated to this work).
- Frontend `npm test`: 2,182 passed, 0 failed, 179 files.
- `npx tsc --noEmit`: 102 errors, the same 102 `main` has; none in files this
  branch touches.
- `npm run build` succeeds. The new view is its own lazy chunk at 14,271 bytes
  raw / 3,959 gzipped; `main.js` is unchanged.

## Deployment

One migration, `20260908_event_webhooks`, adding `webhook_endpoint` and
`webhook_delivery`. The delivery worker starts with the API process and is
gated on `WEBHOOK_DELIVERY_ENABLED` (default on); with it off, events are still
recorded in the outbox and nothing is posted.

> [!NOTE]
> **[Medium]**
> New outbound egress surface (signed, at-least-once) touching the approval/policy/gateway hot paths. The budget-event commit gap, circuit-breaker probe issues, and synchronous `policy.denied` emit are all fixed; remaining items are LOW concerns (SSRF guard off by default, a fire-and-forget enqueue crash window) plus documentation cross-references.
>
> **Overview**
> Migrates outbound webhooks onto a signed, transactional-outbox delivery path with seven v1 governance events, endpoint CRUD + console, a compatibility shim for the legacy approval webhook, and a documented SSRF guard that is off by default.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 6e686599. Updates automatically on new commits.</sup>


d-mo:
Merged `github/main` (`a746dc11`) so this branch parents `20260908_event_webhooks` on `20260908_approval_park` (the new single Alembic head after #510).

The previously skipped **[Performance]** item on `policy.denied` is fixed in `6e686599`: deny no longer runs `enqueue_event_detached` (select + count + insert + commit) on the caller thread or event loop. `_log_policy_decision_async` now snapshots the payload and `submit_off_loop`s the emit onto the shared DB thread pool. Tests cover that deny schedules the work instead of running it inline, and that allow/approval do not enqueue.

PR #510, opened 2026-09-08
Approvals on a human timescale: park the execution, resume on the decision

d-mo:
# Approvals on a human timescale: park the execution, resume on the decision

## Why

Dogfood run on staging, 2026-09-08. Preset 006 raised one batched `ask_user` for four Go advisories
at 00:47:13 with `expires_at` at +5 minutes. The agent container gave up at about 2.5 minutes and
exited 0, the platform marked the execution FAILED `cra_result_missing` at 00:49:47, and a human
approved at 00:50:08: 175 seconds after the ask, 27 seconds too late. A correct human risk
acceptance attached to no evidence pack at all.

Five minutes is not a bad constant, it is a structural mismatch. The platform held a container, a
runner slot, a wall-clock budget and an in-process poll loop open while a person was expected to
make a compliance decision. In that shape anything longer than a coffee break is unaffordable, so
the default is short, so real decisions do not fit. The fix is not a bigger number, it is to stop
burning the run.

## What this does

**1. The approval window is a first-class setting.** `resolve_approval_window()` picks the most
specific value: `timeout_seconds` passed to `ask_user`/`request_approval`, else the flow's
`approval_window_seconds`, else the approval workflow timeout, else the unchanged 5 minute default.
Everything is clamped to the deployment ceiling (30 days) and to the per-account cap in
`meta_data.approval_window_max_seconds`, which may only tighten. An existing workflow with a short
timeout is never lengthened. Reminders re-notify through existing preferences at 50 and 90 percent
of the window.

**2. The execution parks instead of burning.** After 90 seconds of in-process waiting
(`APPROVAL_PARK_AFTER_SECONDS`) on a window longer than that, the tool returns a
`parked_for_human` result telling the agent to stop rather than poll, the orchestrator moves the
execution to the new non-terminal `WAITING_FOR_HUMAN` status, and the container and runner are
released. Park is a three-process handshake over columns the console needs anyway: the approval
path writes `park_request_id`, the orchestrator monitor confirms it, the decision path claims it.
Isolated publication is skipped while parked, because publishing now would ship exactly the change
nobody has approved.

**3. The decision resumes the run.** Every resolution path (console, mobile, API, tokenized public
link) funnels through `update_approval_request`, which claims the parked row with a single
conditional UPDATE (a duplicate decision claims zero rows and is a harmless no-op) and starts a new
execution carrying the existing `_resume` continuation contract from #490/#419, so the workspace
snapshot, the CLI session pack and the model pin come back. The answer travels three ways:
`payload.answers[<request_id>]` for container agents that cannot resume, an `_answers_prompt` block
appended to the prompt (same mechanism as `_feedback_prompt`), and the request id so the model can
bind it to what it asked. Expiry is a resume too, with an `expired` answer, so the agent finishes
and writes its artifact instead of the platform inventing `cra_result_missing`.

**4. The flow's compute budget pauses while parked.** `TimeoutBudget` gained `consumed_seconds`;
the resumed run's budget is the flow budget minus the chain's `parked_compute_seconds`, floored at
the minimum. Human waiting time never counts. A flow with a 2 hour budget and a 3 day window gets 2
hours of agent wall clock, whenever it is spent.

**5. Console and presets.** `WAITING_FOR_HUMAN` is a warning-variant chip in the executions list
(a status, not a spinner: nothing is spinning) with a row title naming who and until when, and the
execution page shows "waiting for X, since Y, expires Z" plus the question. The flow form takes the
window as an amount plus a unit (minutes, hours, days), because compliance windows are set in days
and nobody wants to type 259200. Presets 006 and 014 set 3 days.

## Honest limitation

A resumed native session receives the answer as its next turn, not literally as the return value of
the suspended tool call: no harness exposes "inject a tool result into a session that was killed".
The prompt block names the `request_id` and the question so the model can bind it. This is stated in
`docs/architecture/approvals.md` and in the preset guide.

## Migration

`20260908_approval_park` adds `flow.approval_window_seconds` and five `flow_execution` park columns
plus two indexes. Downgrade drops exactly what upgrade added; verified against a real Postgres
database (`upgrade`, `downgrade`, `upgrade`) and asserted in the test.

## Tests

* `backend/tests/test_approval_park.py`: 67 tests covering window resolution and clamping, park
  decision, pending payload, answer projection, resume details, reminders, the status transition
  SQL (rendered with literal binds, so the predicates are asserted rather than object reprs),
  resume on decision, duplicate decision, budget pause, parked being non terminal, and migration
  up/down symmetry.
* Preset tests pin the 3 day window on 006 and 014 and pin that the window outlives the run's own
  compute budget, which is the property staging violated.
* Frontend: 28 new tests over the waiting chip, the execution page waiting line and the flow form
  window control.
* Full frontend suite 2074 passed, 0 failed. `tsc --noEmit` error count unchanged against the base
  commit (97 both sides, all pre-existing). Bundles: `preloop-flow-form` +3.6 kB,
  `flow-execution-view` +1.25 kB, `flow-executions-view` +0.11 kB, `main` unchanged.

No changes to the answer *shape*: `question-answer-panel.ts` is untouched and the payload is
additive, so the sibling structured-questions work merges cleanly.

> [!NOTE]
> **Low**
> No security vulnerabilities; the park/resume design is sound. All previously raised HIGH, MEDIUM, and LOW items (crash-stuck `RESUMING`, duplicate-resume window on failed dispatch, preset 006 waiver instructions, `_release_parked_executions` guard, and the parked row staying `RESUMING` after a successful resume) have been addressed.
>
> **Overview**
> Makes the approval window a first-class per-flow setting and replaces "hold a container while a human decides" with a park/resume handshake: the execution moves to the non-terminal `WAITING_FOR_HUMAN` status, the runtime is released, and the decision (or expiry) resumes it via the existing `_resume` contract while the compute budget pauses. This update adds crash recovery for the transient `RESUMING` state, prevents a double-resume after a failed dispatch, and closes the parked parent once the resume child reaches a terminal state.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit d2dd8bce. Updates automatically on new commits.</sup>


d-mo:
Addressed the HIGH and MEDIUM review notes on `a95f88d8` (also merged `github/main`).

**HIGH — RESUMING row stranded on crash.** `claim_parked_for_resume` now stamps the ordinary claim lease (`orchestrator_heartbeat_at`, same stale window as other claims). The parked sweep reclaims stale `RESUMING` rows that have no `resume_execution_id`, returns them to `WAITING_FOR_HUMAN`, and retries resume. Regression: claim into `RESUMING`, skip create, run sweep → a resume execution is created.

**MEDIUM — failed dispatch can double-run.** The PENDING resume insert and `resume_execution_id` link commit in one transaction. Dispatch runs after that commit. A failed dispatch no longer releases the claim, so the sweep will not create a second execution. Regression: dispatch fails after the insert; sweep must not start another run.

**Also in this push:** preset 006 waiver collection now looks for the parked `_answers_prompt` block, and `_release_parked_executions` looks up by `park_request_id` instead of guarding on `ApprovalRequest.execution_id`.

Issue #652, opened 2026-09-15
feat(sessions): add a ranked keyword search endpoint over session content


d-mo:
Child of #624.

## Summary

Add a ranked keyword search endpoint over the session corpus: a POST endpoint taking a query, filters and paging, returning sessions ordered by relevance with snippets pointing at the matching turn.

## Why

The corpus is useless without a query surface. The existing content search is unranked and reachable only per session or account wide behind a cost permission, and it orders by timestamp, which is the wrong answer to "where did an agent do this". Ranking matters more than it sounds: a session that matches in five places is almost always the one you want, and a timestamp ordering buries it.

POST rather than GET is a deliberate departure from the two existing GET search endpoints, so query text stays out of access logs. Query text here is whatever an operator is looking for in their own agent transcripts, which is exactly the kind of string that should not end up in a proxy log.

## Scope

- A POST search endpoint over runtime sessions, taking a query string, a mode, filters, a limit capped at a documented maximum, an offset, a maximum number of snippets per session, and a flag for whether snippet text is returned at all.
- Filters over the denormalised columns the corpus carries: time range, model, provider, principal, api key, flow, source kind.
- Query semantics use the database's web search query parsing, so quoted phrases, alternation and exclusion behave the way a user expects, over the same text configuration the corpus is indexed with.
- Ranking: rank by relevance, and fuse chunk scores into one session score with a documented bonus for a session matching in several distinct chunks. The constants are documented as tunable and unvalidated.
- Snippets generated by the database's headline function rather than a substring window, each with its timestamp and the identifier needed to open the session at that turn.
- Response carries the query, the mode, a degraded block, an indexed through marker, the total session count, the elapsed time and the results.
- Unknown filter keys are a validation error rather than being ignored.
- Modes other than keyword are accepted and answered with keyword results plus a degraded marker saying semantic is not enabled, so the endpoint contract does not change when semantic lands in #657.
- Account scoping enforced in the query itself, not in the serializer.

## Not in scope

Vector search and fusion, which are #657. The console, which is #656. The MCP tool, which is #658. Backfill, which is #653. Saved searches.

## File pointers

Verified on main `67b8c90d`.

- The corpus being queried: from #649.
- The existing approach and its limits, for comparison: `backend/preloop/models/crud/gateway_usage_search_document.py:130-140` (the full text predicate), `:181-184` (ordering by timestamp, not rank).
[...]

PR #724, opened 2026-09-15
feat(cli): add preloop sessions search

### LINKED ISSUE #659: feat(cli): add preloop sessions search
Child of #624.

## Summary

Add `preloop sessions search` to the CLI: query session content from a terminal, with human readable output by default and machine readable output on request.

## Why

Self hosted operators live in a terminal and in scripts, and the console is not always where the question gets asked. A search that exists only in a browser cannot be piped into a report, cannot be run over ssh on the box that is misbehaving, and cannot be dropped into a runbook.

It is a thin client over the endpoint, which is what makes it cheap and what keeps it honest: no query logic lives here, so the CLI cannot drift from the console.

## Scope

- A search subcommand under the sessions command group, taking the query, a time range, a limit and a mode.
- Default output is readable: session reference, timestamp, snippet, match reason, one result per block.
- A json output flag emitting the endpoint response shape unchanged, for piping.
- Degraded markers are printed to the error stream so a piped json payload stays clean while a human still sees that coverage was partial.
- Exit codes distinguish results found, no results and an error, so scripts can branch.
- Paging through more results than the page size, either by a flag or by following the endpoint's cursor.
- Help text and a documentation example.

## Not in scope

Any query logic in the CLI. Interactive browsing of a session. Output formats other than the readable one and json.

## File pointers

Verified on main `67b8c90d`.

- The endpoint being called, its request shape and its degraded block: from #652.
- The session identifiers rendered in output come from the runtime session model surface: `backend/preloop/models/crud/runtime_session.py:482-491` is the identifier filter that the content search replaces, and is the reference for which identifiers users already type.

## Acceptance criteria

- The subcommand appears in help with its flags documented.
- A query prints readable blocks, one per result, asserted on a fixture response.
- The json flag emits the endpoint payload unchanged, asserted by comparison.
- Degraded markers appear on the error stream and not in the json payload.
- Exit codes differ for results found, no results and an endpoint error, one assertion each.
- Paging retrieves the second page, asserted on the requests made.
- An endpoint error prints a readable message rather than a traceback.

## Dependencies

#652.

d-mo:
## Summary

`preloop sessions search` queries recorded session content from a terminal, over
the ranked keyword endpoint added in #652 (`POST /api/v1/runtime-sessions/search`).

It is a thin client on purpose: no query is parsed, no ranking is computed and no
filter is applied in the CLI, so the same words return the same answer here and in
the console. What it adds is what a terminal needs and a browser cannot give: an
answer that can be piped into a report, read over ssh on the box that is
misbehaving, and branched on in a script.

## Changes

- `cli/internal/cmd/sessions.go` (new): the `sessions` command group and its
  `search` subcommand. Flags: `--mode` (keyword, semantic, hybrid), `--from` /
  `--to` (a calendar day, read as UTC midnight, or an RFC 3339 instant; the
  endpoint rejects naive values), `--limit`, `--page-size` (at most 50, the
  server's own cap) and `--json`.
- Readable output is one block per session: the session id and the source
  type/id operators already type, the reference and title when set, when it ran,
  the match reason (matching chunk count, fused score, best chunk rank) and each
  matching turn with its timestamp, source kind, role, chunk index and rank.
  `<mark>` headline markup from the server is stripped in this view.
- `--json` writes each response page exactly as the endpoint sent it, one
  document per page, so it can be piped into `jq`. Coverage notes go to the
  error stream instead: a degraded ranking mode and its reason codes, the mode
  that actually ran, the indexed through marker, the result count and a hint
  when more results exist. A piped payload therefore stays a clean document while
  a human still sees that coverage was partial.
- Paging: `--limit` is the total to retrieve, `--page-size` how many per request.
  A limit above the page size is paged by offset (the endpoint pages by offset,
  not by cursor), stopping on the first of enough results, a short page, or the
  total the server reported.
- Exit status: 0 when something matched, 2 when the search ran and matched
  nothing, 1 when it could not be answered. `cli/internal/cmd/exitcode.go` gains
  `exitCodeError` so a command can choose a status for its own outcome, and
  `ProcessExitCode` honours it. The command reports its own failures once and
  silences cobra afterwards, so an empty answer is not dressed as an error and a
  500 with a stack trace in its body becomes one sentence naming the status.
- Refusals use the server's own `detail`, with the permission named on 403
  (`view_runtime_sessions`) and a 404 read as "this deployment does not have the
  endpoint".
- Docs: a `### Session search` section in `cli/README.md` with examples, plus the
  command help text (`Examples:` block and the exit status contract), and a
  CHANGELOG entry.

## Tests

`cli/internal/cmd/sessions_test.go` drives the command through the root against an
`httptest` endpoint that records every request, one test per acceptance
criterion:

- the subcommand appears in the root help, under `sessions`, and its help
  documents every flag;
- a query prints one readable block per result, asserted on a two result fixture
  (identifiers, timestamps, match reason, snippet text without markup, and the
  "no snippet text" line for a metadata only chunk);
- `--json` output equals the fixture payload byte for byte, and again after
  decoding both sides;
- degraded markers (`semantic_not_enabled`, its detail sentence, the effective
  mode) appear on stderr and never in the payload, checked for both output modes;
- exit codes: 0 with results, 2 with none (stdout empty, no `Error:` on stderr),
  1 on an endpoint failure;
- paging retrieves the second page, asserted on the requests made (limit 2 offset
  0, then limit 1 offset 2) and on three blocks printed; a full page that
  exhausts the total makes no second request;
- an endpoint error prints a sentence, asserted on a body that is a Python
  traceback: neither stream carries it and the failure is reported once;
- plus the time range reaching the filter block as timezone explicit instants,
  local flag rejections making no request, and unquoted words joining into one
  query.

Counts (`cli/internal/cmd`, the only package touched):

- focused: `go test ./internal/cmd/ -run TestSessions` 15 top level tests
  (4 subtests) pass; `-run 'TestProcessExitCode|TestWrapProcessExit'` 4 top level
  tests (5 subtests) pass, including the new
  `TestProcessExitCodeHonoursACommandChosenStatus`.
- whole package: `go test ./internal/cmd/` 843 pass, 0 fail, 6 skipped (Docker
  and platform gated, unchanged), 454 subtests. `go build ./...`, `go vet
  ./internal/cmd/` and `gofmt` clean; `golangci-lint run ./internal/cmd/...`
  reports no issue in the files touched here (its 37 pre-existing findings in
  other files are unchanged). `pre-commit run --files <changed files>` passes; no
  Python changed, so `openapi.yaml` is untouched.

## Not in scope

No query logic in the CLI, no interactive browsing of a session, and no output
format other than the readable one and json. The corpus filters beyond the time
range (model, provider, principal, api key, flow, source kind) and the snippet
controls (`max_snippets_per_session`, `include_snippet_text`) stay with the
endpoint until a terminal asks for them.

Closes #659

> [!NOTE]
> **Low risk**
> A read-only CLI client over an already-merged endpoint; both findings from the first review (sub-second bound truncation and untested refusal branches) were fixed in bc878de1, and no security or paging defects were found.
>
> **Overview**
> Adds `preloop sessions search` (plus a `sessions` command group) as a thin terminal client over `POST /api/v1/runtime-sessions/search`, with readable blocks, byte-faithful `--json` paging output, coverage notes on stderr, and a 0/2/1 exit-status contract for matched/no-match/failure. Extends the exit-code plumbing so a command can choose the status for its own outcome.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit bc878de. Updates automatically on new commits.</sup>


d-mo:
Both review findings are addressed in bc878de1.

- The sub second bound: `--to 2026-09-15T12:00:00.250Z` is now formatted with `time.RFC3339Nano`, so the fraction reaches the endpoint instead of being truncated. Whole second inputs are byte identical to before. New test `TestSessionsSearchKeepsASubSecondBound`.
- Refusal coverage: `TestSessionsSearchStatusesBecomeSentences` is a status to sentence table over `explainSessionSearchError`, covering 401 (points at `preloop login`), 404 with an HTML body (asserts the body is not echoed), 422 with a FastAPI field error list (asserts each field is named) and a generic status. Each case also asserts exit 1 and an empty stdout.

Focused tests: `go test ./internal/cmd/ -run TestSessions` 15 top level tests and 8 subtests pass. Whole package: 845 pass, 0 fail, 6 skipped (Docker and platform gated), 458 subtests.

PR #770, opened 2026-09-16
fix(console): upgrade modal only on user actions; plan picker defaults; Free hides the Stripe portal

d-mo:
## Summary

A brand new Free account met the paywall dialog on the first screen it ever
loaded: the dashboard reads the model price override list, the plan gates that
endpoint with a 402, and every 402 opened the dialog. This makes the dialog an
answer to a click and nothing else, and fixes the three plan controls that sent
readers to the wrong place: the dialog's hardcoded checkout, the plan picker
that opened on the plan the account was already on, and a "Manage in Stripe"
button that is permanently disabled on Free.

Phase 1a of the console plan UX work. Phase 1b (a dedicated
`/console/settings/plan` page reusing the public pricing cards, and the
emergency controls on their own page) is stacked on this branch.

## Founder decisions applied

- (5) The paywall modal appears only on a user action, never from a background
  fetch. `fetchWithAuth` takes `passive`, which suppresses the dialog (402 and
  429) and returns the response so the caller can render nothing. Applied to
  every boot-time and background read that can hit a gated endpoint.
- (2) Upgrades and downgrades are unchanged here: this change only decides
  which plan the picker opens on, not what a change costs or when it applies.

## Changes

- `frontend/src/api.ts`: `AuthFetchOptions` adds `passive`. The flag is
  stripped before `fetch`, coalesces passive and active reads of the same URL
  separately (so a background read cannot swallow the dialog a click deserves,
  and vice versa), and suppresses both the 402 upgrade dialog and the 429
  dialog for passive calls. `getModelPriceOverrides({ passive: true })`
  resolves to an empty list on 402 instead of throwing.
- `frontend/src/utils/attention-data.ts`: the price override leg of the
  dashboard `Promise.allSettled` is passive.
- `frontend/src/views/authed/cost-view.ts`: the page-load price override read
  is passive. The Reconciliation tab keeps its dialog: opening the tab is a
  click.
- `frontend/src/utils/premium-features.ts` (new): one place for the feature
  labels, the feature to capability mapping (`session_optimization`,
  `replay_verification` and `session_titles` are all sold as
  `ai_optimization`), and the plan page URL. Phase 1b flips the path constant.
- `frontend/src/views/authed/console-shell.ts`: the dialog no longer starts a
  checkout for a hardcoded plan. "View plans" closes it and opens the plan
  page; "Upgrade now" closes it and opens the same page with
  `?feature=<capability>` so the page can preselect the cheapest plan that
  unlocks the refused feature. In this PR the plan page is the account view;
  Phase 1b points the constant at `/console/settings/plan`.
- `frontend/src/components/billing-plan-comparison.ts`: default billing period
  is annual; the "Change to" list excludes the plan the account is already on
  (a Free account used to open on "Free"), except where the same plan on the
  other billing period is a genuine change; the default target is the cheapest
  plan above the current one; `?plan=`, `?interval=` and `?feature=` are
  honoured and open the picker; options are selected with `?selected=` because
  Lit set `.value` before the options existed, which displayed the first entry
  whatever the selection was; a public `openPicker()` for the account page.
- `frontend/src/views/authed/settings/account-view.ts`: with no subscription,
  "Manage in Stripe" is replaced by "Choose a plan", which opens the picker and
  scrolls to it.

## Tests

Frontend, `@web/test-runner` (chromium):

- Touched files: 263 passed, 0 failed across `api.test.ts`,
  `premium-features.test.ts`, `attention-data.test.ts`,
  `billing-plan-comparison.test.ts`, `console-shell.test.ts`,
  `attention-view.test.ts`, `cost-view.test.ts`, `account-view.test.ts`.
- Full suite: 207 files, 2821 passed, 0 failed.
- `npm run build`: passes.
- `pre-commit run --files <changed files>`: passes.

New tests include: the attention view mounted with a 402 raises no
`show-upgrade-modal` event and still renders; the same for the cost view and
for `loadAttentionInputs`; an active 402 still opens the dialog; a passive read
and an active read of the same URL are not coalesced; the flag never reaches
`fetch`; where no plan gate exists (the self-hosted default, 200 answers) a
passive call behaves exactly like any other call; the picker opens annual, on
the next tier up, never on Free, never on the current plan; `?plan=`,
`?interval=` and `?feature=` are honoured, including the case where no plan
proves it includes the feature; Free sees "Choose a plan" and it opens the
picker.

## Operator notes

- No backend, schema or plugin change. No Stripe call was made from this
  machine; every billing endpoint in the tests is a stub.
- Self-hosted Preloop is unaffected: without the billing plugin no endpoint
  answers 402, and `passive` changes nothing about the request that is sent.
  Covered by a test.
- The dialog copy changed. It now points at the plan page instead of promising
  a one-click purchase.

## Not in scope

- `/console/settings/plan` and `/console/settings/emergency` (Phase 1b, stacked
  on this branch).
- The plan page's per-card CTAs, the quote step and the checkout path for Free
  accounts (Phase 1b).
- Anything on the backend: capability names, gate placement and the 402 body
  are unchanged.

> [!NOTE]
> **Low Risk**
> Frontend-only billing UX with no backend, schema or Stripe surface; the one defect found in the first review pass (a raw NUL byte in the passive coalesce key in `frontend/src/api.ts`) was fixed in `83ba919d` — the key is now built by an exported `coalesceKey()` with a printable `passive|` separator and a regression test asserting printable-only keys — so no active findings remain.
>
> **Overview**
> The upgrade dialog now appears only as an answer to a user action: background and boot-time reads mark their fetches `passive`, which suppresses the 402/429 dialogs and lets callers render nothing. The dialog points at the plan page instead of a hardcoded checkout, the plan picker defaults to annual, opens on the next tier up (never the current plan) and honours `?plan=`/`?interval=`/`?feature=`, and Free accounts get "Choose a plan" instead of a permanently disabled Stripe button. Phase 1a of the console plan UX work; no backend changes.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 83ba919d. Updates automatically on new commits.</sup>


d-mo:
## NUL byte fixed (`83ba919d`)

Thanks, that was a real defect and a genuinely invisible one. Confirmed and fixed.

### What was there
One byte, `frontend/src/api.ts` line 385, column 43 (byte offset 12785 in the file):

```
$ sed -n '385p' frontend/src/api.ts | od -c
0000040    ?       `   p   a   s   s   i   v   e  \0   $   {   u   r   l
```

Exactly one NUL in the file, on the line this PR added, and it was the only control character anywhere in the PR. I scanned both the full content of all 15 changed files and the raw added lines of `git diff origin/main...HEAD` for `[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]`, plus the usual invisible Unicode (zero width space, joiners, bidi marks, BOM, non breaking space): the NUL at `api.ts:385` was the single hit. `grep -rlP '[\x00-\x08\x0b\x0c\x0e-\x1f]' frontend/src` now comes back empty.

### What replaces it
Not a space. A space is a legal character inside a URL string, and more to the point an invisible separator is exactly how this happened, so the separator is now a visible pipe and it lives in a named function with a comment that says why:

```ts
export function coalesceKey(url: string, passive?: boolean): string {
  return passive === true ? `passive|${url}` : url;
}
```

`fetchWithAuth` calls `coalesceKey(url, options.passive)`. No request URL starts with a pipe, so a passive key can never collide with an active one, and the passive/active split the tests assert is unchanged. Behaviour is identical: the only difference is that the separator is now something a human and a text tool can both see.

### Tests
Two new cases in `src/api.test.ts`, in the existing passive block:

- `namespaces the passive key with printable characters only`: asserts the key equals `passive|/api/v1/billing/cost/pricing-overrides`, that `/[\x00-\x1f\x7f]/` does not match it, that it is entirely `[\x20-\x7e]`, and that the active key stays the bare URL. I verified this fails as intended by temporarily putting the NUL back: `namespaces the passive key with printable characters only` fails, so the byte cannot silently return.
- `does not coalesce passive reads of two different gated features`: two concurrent passive GETs to two different gated endpoints produce two `fetch` calls and each caller reads its own body.

`src/api.test.ts` 47 passed, 0 failed. All 8 test files touched by this PR: 266 passed, 0 failed. `npm run build` succeeds. `pre-commit run --files frontend/src/api.ts frontend/src/api.test.ts` passes (prettier reformatted the new test block; result re-tested).

`grep -rn "startCheckout" frontend/src` now prints line matches for `api.ts` instead of "binary file matches", which is the observable symptom you described going away.

### On the low priority item (`startCheckout`)
Keeping it, deliberately. Phase 1b (the stacked plan page PR) reuses it: a Free account with no subscription goes straight to Stripe checkout via `startCheckout`, with no quote step, because there is no subscription to switch. Deleting it here would only mean re-adding it in the next PR. Noted here as you asked rather than left silent.

New head: `83ba919d`. Pushed as a fast-forward on top of `419ab79d`, no force push. No inline review threads exist on this PR to resolve (the inline payload was rejected at review time), so this comment is the reply to the summary.

d-mo:
Stacked follow-up: #774 (the plan page and the emergency page) is built on this branch. Merge this one first.

d-mo:
Merged main after #769/#772; conflicts in `frontend/src/views/authed/console-shell.ts` and `frontend/src/api.test.ts`, resolved by carrying main's three new gate labels (`price_overrides`, `ai_optimization`, `analytics_window_days`) into `utils/premium-features.ts`, where this branch had already moved that table, and keeping both sides' imports in the test; tests: 2876 passed / 0 failed (full `npm test`, 211 files), 277 passed / 0 failed on the 8 files this PR touches, 70 passed / 0 failed on `api.test.ts` plus the two new nudge components, `npm run build` and `npm run typecheck` clean.

Two follow-on changes the merge itself required, since neither side is wrong on its own:

- `getUsageNudges` is now a passive request. The nudge banner loads on every console page without anyone asking for it, so a 429 there would have raised the rate-limit dialog over whatever the reader was doing, which is the behaviour this PR exists to remove. Covered by a new case in `api.test.ts`.
- `frontend/tests/e2e/upgrade-flow.spec.ts` asserted the old dialog (start a checkout for a hardcoded plan). It now asserts the new one: the dialog closes, the plan page opens with `?feature=ai_optimization`, and no checkout request is made.

PR #772, opened 2026-09-16
feat(billing): plan-fit enforcer inputs, analytics window feature key

d-mo:
## Summary

Two primitives the enterprise billing plugin needs for the subscription
lifecycle work, and nothing else. Open source behaviour is unchanged: the new
CRUD methods have no caller in core, and the new plan feature key is optional
with a fallback to today's value.

## Founder decisions applied

* Cancellation returns an account to Free and the account is actively
  reconciled to the Free limits. The ordering, the write and the audit row for
  that reconciliation live here; the policy that decides whether to run it
  stays in the plugin.
* Free shows 90 days of analytics at read time while physical retention is
  unchanged. `analytics_window_days` is a read-time window only, and the
  purger keeps reading `retention_days`.

## Changes

* `models/crud/billing.py`
  * `plan_fit_candidates(db, account_id)`: an account's active members and
    agents, most keepable first. The owner (`account.primary_user_id`) is
    always first, then members by `User.last_login` (never logged in sorts
    last), then `created_at` and id so the order is total and stable. Agents
    are ordered by `ManagedAgent.last_seen_at`, the timestamp enrollment and
    gateway traffic stamp, then `created_at` and id. Read only.
  * `enforce_plan_fit(db, *, account_id, max_users, max_agents, reason,
    operation_key, payload)`: trims one account to a pair of caps in one
    transaction under the account lock. Surplus members are deactivated
    (`is_active=False`), surplus agents are suspended (`lifecycle_state`,
    `lifecycle_reason`, control session cleared) exactly as an operator pause
    does. A cap of -1 leaves that half alone. Two layers of idempotency: a
    completed operation key is reported back untouched, and an account already
    inside its caps writes nothing. One `billing_operation` row records every
    enforcement. It decides nothing: the caps arrive as numbers.
* `models/crud/history_policy.py`: `ANALYTICS_WINDOW_KEY` and
  `analytics_read_window(features)`, which falls back to
  `analytics_plan_retention` when a plan does not state a window.
  `analytics_plan_retention` is documented as the storage promise and the
  purger's only input.
* `models/schemas/plan.py`: optional `analytics_window_days` field.
* `services/analytics_history.py`: the reporting window and the storage floor
  are now two questions. `analytics_storage_policy` (a new optional plugin
  service) answers what a plan KEEPS; without it the storage path falls back
  to the reporting policy, which is exactly today's behaviour. Policy answers
  are validated once in `_policy_days`, and a broken answer raises 503 instead
  of silently shortening a window.

## Tests

Run against Postgres, with `DATABASE_URL` set.

* `tests/models/crud/test_plan_fit.py` (new): 12 tests. Candidate ordering
  (owner first without a login, last login order with never-logged-in last,
  agents by `last_seen_at`, inactive rows excluded) and enforcement (surplus
  members deactivated and the owner kept, agents suspended and disconnected,
  within limits writes nothing, -1 trims nothing, the audit row, a replayed
  key is `already_applied` and does not undo a manual reactivation).
* `tests/models/crud/test_history_policy.py`: 27 tests, 2 new (the window
  defaults to storage and nonsense is ignored; the read window is invisible to
  the purge input).
* `tests/services/test_analytics_history.py`: 46 tests, 4 new (a shorter
  display window does not shorten storage; storage falls back to the reporting
  policy; open source has no storage policy; broken policy values raise 503).
* `tests/models/test_plan_schema.py`: 18 tests, 1 new.

Totals: 103 passed across the four touched files. A wider
`tests/models tests/services` run has 18 failures that are present on this
branch and on its merge base alike (time and environment sensitive tests,
none in the modules touched here).

`pre-commit run --files` passed on every file in this PR, including the
OpenAPI generation hook: `openapi.yaml` is unchanged because `PlanFeatures` is
not referenced by any router.

## Operator notes

* Nothing in this PR runs on its own. Without the enterprise billing plugin
  there is no caller for `enforce_plan_fit` and no
  `analytics_storage_policy`, so a Preloop OSS install behaves exactly as it
  did before.
* `analytics_window_days` is a read-time window. It never deletes anything,
  and raising it back on an upgrade immediately reveals rows that were always
  there.

## Not in scope

* The policy that decides when an account is trimmed, the owner notification
  and the plan catalog key itself: those live in the enterprise billing
  plugin merge request (`billing: cancellation notifies admins and reconciles
  to Free; upgrades immediate with a fresh cycle; downgrades at period end;
  duplicate checkout guard`).
* The retention purger and `retention_days` are untouched.

> [!NOTE]
> **Low Risk**
> No security issues and OSS behaviour is verifiably unchanged; the three MEDIUM concurrency/correctness points from the first review (idempotency check outside the account lock, owner deactivation at `max_users=0`, and the relaxed 183→90 floor on the purge-path fallback) are all fixed in `8cc1079` with targeted regression tests.
>
> **Overview**
> Adds the persistence primitives the enterprise billing plugin needs for subscription-lifecycle reconciliation: `plan_fit_candidates` / `enforce_plan_fit` (ordered, lock-serialized, audit-logged trimming of members and agents to plan caps, never the owner) and a read-time `analytics_window_days` plan key whose storage path falls back to today's reporting policy but stays held to the stricter 183-day purge floor. Core ships no callers, so open-source installs behave exactly as before.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 8cc1079. Updates automatically on new commits.</sup>


d-mo:
## Follow-up to the three MEDIUM review threads, after this PR was merged

This PR was merged (2ae5fd5) about nine minutes after `8cc10792` landed, so the round-1 answers to all three MEDIUM threads are in `main`. Reviewing them again against the review text, two were under-specified and one was answered in a way that traded the reported bug for a different one. The corrections are on `feat/billing-lifecycle` at **f36199f5** (the branch was auto-deleted on merge; I re-created it by pushing). **There is no open PR for that commit yet**, since this PR set was authorised for one core PR only. Happy to open a follow-up PR on request.

### 1. Lock before the idempotency read (`billing.py:289-294`)

Fixed in `8cc10792` and merged: `lock_account` now precedes the completed-key read, matching `claim_operation`, with a `db.commit()` before returning `already_applied` so the lock is released.

What was missing is the test. `f36199f5` adds `TestEnforcePlanFitRace::test_two_sessions_with_one_key_apply_once_and_replay_once`: two sessions on independent connections (committed rows, cleaned up after) enforce the same `operation_key` at once. The loser blocks on the account lock and reads the winner's completed row, so the pair is exactly one `applied` and one `already_applied`, one `billing_operation` row, and no `IntegrityError`.

It is a real regression test, not a description of one: with the check moved back before the lock, the loser reports `within_limits` and the test fails with `['applied', 'within_limits'] != ['already_applied', 'applied']`. Verified by reverting the ordering locally.

### 2. `max_users=0` and the owner (`billing.py:294`)

`8cc10792` filtered the owner out of the surplus set. `f36199f5` additionally clamps a finite seat cap below one to one, so the guarantee holds in two independent ways: the clamp means the owner is never sliced into the surplus, and the filter means the "never the owner" docstring does not depend on the owner sorting first in `plan_fit_candidates`. Your recommendation offered either; the consequence of getting this wrong is an account with no administrator and no way to reach the upgrade that would undo the trim, so both are cheap.

The point about the silent test is also fixed. `8cc10792` added a separate test and left the parametrized case asserting nothing; `f36199f5` makes `test_an_unlimited_cap_trims_nothing_on_its_side` assert in every case, including `(0, -1)`, that the owner is still active and is absent from the returned surplus.

### 3. The policy floor and the purge path (`analytics_history.py:30,48,67-78`)

This is the one I changed course on. `8cc10792` kept the reporting-policy fallback in `_configured_storage_days` and validated it against a new 183-day storage bound, so a build publishing only `analytics_history_policy` got a 503 on the purge path. That stops the bad purge, but it still lets a read-time answer decide a purge question, and it converts an older plugin build from "purges correctly" into "purge job fails".

`f36199f5` removes the fallback entirely: only `analytics_storage_policy` can answer the storage question. A build that publishes only the reporting policy now contributes nothing to the purge floor, which stays exactly where it already was (`longest_history_promise` over the account's stored and persisted promises), rather than being fed a display window or failing closed.

The 183-day bound is now a floor rather than a veto: a storage answer below it is raised to 183 instead of raising. Keeping more rows than a plan promises is always safe; deleting to a display window is not.

Founder decision applied: physical retention is unchanged, and only the read-time analytics window may be 90 for Free.

Tests (`tests/services/test_analytics_history.py`):
* `test_a_ninety_day_policy_shows_ninety_days_and_purges_at_the_storage_floor`: a policy answering 90 yields a 90-day read window and a 183-day purge floor.
* `test_a_reporting_only_plugin_leaves_the_purge_floor_where_it_was`: a plugin publishing only the read policy leaves the purge floor at its previous value (365 here), with `_configured_storage_days` returning `None`.
* `test_a_reporting_only_plugin_never_shortens_an_unpromised_account`: the same with no promise on the account, so the 90-day window cannot become the delete boundary for the accounts the review identified as exposed.
* `test_a_storage_policy_longer_than_the_floor_is_honoured`: 183 is a floor, not a cap.

### Tests

* `tests/models/crud/test_plan_fit.py` 14, `tests/services/test_analytics_history.py` 49, `tests/models/crud/test_history_policy.py` 27, `tests/models/test_plan_schema.py` 18: **108 passed**.
* Wider `tests -k "analytics or retention or history or purge"`: **268 passed** (excluding `tests/services/test_disposable_kubernetes_guard.py`, which fails to collect on this machine before and after).
* `pre-commit run --files` passed on all four files, including the OpenAPI hook; `openapi.yaml` unchanged.

The enterprise side is https://gitlab.spacecode.ai/spacecode/preloop-ee/-/merge_requests/77, still open, with its own HIGH finding fixed and all discussions resolved. It passes against core `main` at 36c59ac6: 843 passed, 3 pre-existing failures.

PR #859, opened 2026-09-21
fix(openclaw-plugin): honour config.enabled=false by registering nothing (0.3.1)

### LINKED ISSUE #857: fix(openclaw-plugin): `config.enabled=false` is advertised as "installed but inactive" but the runtime never checks it
`runtime-plugins/openclaw-preloop/openclaw.plugin.json` declares
`config.enabled` ("Set to false to keep the plugin installed but inactive"),
onboarding writes `"enabled": true` into
`plugins.entries.preloop-plugin.config`, and the README shows it. The runtime
declares the field on `ControlConfig` (`src/index.ts:10`) and never reads it:
`register()` configures the instance from `api.pluginConfig`, then
unconditionally schedules `instance.start()` (which opens the reconnecting
Agent Control WebSocket) and registers the `before_tool_call` hook. Only
`tool_approval_enabled` is enforced (`src/index.ts:427`), and it disables
approval gating, not the control channel.

Consequence: an operator who sets `enabled: false` expecting the plugin to
go quiet still has a live control channel accepting `send_message`,
transcript and interrupt commands from the configured control plane.

ClawHub's LLM review flags exactly this on 0.3.0 (and, on inspection, did on
0.2.1): "an advertised disable setting does not appear to be enforced by the
shipped runtime", verdict `suspicious`, which is what the catalog shows for
the package. The static, VirusTotal and SkillSpector scans are clean.

Fix: in `register()`, when `pluginConfig.enabled === false`, log one line
("Preloop plugin installed but disabled by config.enabled=false; no control
channel, no tool-call hook") and return before registering any hook or
starting the instance. Also honour a runtime flip: if the host re-registers
with `enabled: false`, stop. Document the three switches side by side in the
README table (`enabled` kills everything; `tool_approval_enabled` only the
gate; uninstall removes the plugin). Bump `@preloop-ai/openclaw-plugin` to
0.3.1 in both `package.json` and `openclaw.plugin.json` per
`runtime-plugins/PUBLISHING.md`, with a CHANGELOG line for the plugin.

Tests (`runtime-plugins/openclaw-preloop/test/`): `register()` with
`enabled: false` registers no hooks and never calls `start`; with `enabled`
absent or `true`, behaviour is unchanged; with `tool_approval_enabled: false`
and `enabled` absent, the channel still starts (existing semantics).

After merge, publish 0.3.1 to npm and ClawHub through the EE job and confirm
ClawHub's `llmAnalysis.verdict` for 0.3.1 is no longer `suspicious`.

d-mo:
## Summary

The OpenClaw plugin manifest advertises `config.enabled` as a way to keep the package installed but inactive. `register()` declared the field and never read it, so it always scheduled `instance.start()` (the reconnecting Agent Control WebSocket) and always registered `gateway_start`, `gateway_stop`, and `before_tool_call`.

An operator who set `enabled: false` still had a live control channel that accepted `send_message`, transcript, and interrupt commands. ClawHub's review of 0.3.0 (and 0.2.1) flagged this advertised disable setting as unenforced; the other scans were clean.

This change makes `register()` resolve `api.pluginConfig?.enabled`. When the value is exactly `false`, it logs one line and returns before any hook is registered and before `start()` is defined or scheduled. `undefined` and `true` keep the previous behaviour. `tool_approval_enabled: false` remains a gate-only switch and still starts the channel.

README now lists `enabled` as the first configuration key and states that `enabled: false` is the supported pause switch. Plugin version is 0.3.1 in `package.json`, `package-lock.json`, and `openclaw.plugin.json`.

Tests cover `enabled: false` (no `api.on`, no `start`), `enabled: true` and absent (same hooks as before), and `tool_approval_enabled: false` with `enabled` absent (channel still starts, `gateway_start` still registered).

Fixes #857

## Test plan

- [x] `cd runtime-plugins/openclaw-preloop && npm test` (61 pass, 0 fail)
- [ ] Confirm ClawHub review of 0.3.1 after the separate EE publish job (out of scope here)

> [!NOTE]
> **Low Risk**
> A 7-line early-return guard at the plugin's single registration entry point; the strict `=== false` check cannot alter behaviour for existing configs that omit `enabled` or set it `true`, and all 61 tests pass.
>
> **Overview**
> `register()` now honours the manifest-advertised `config.enabled=false` switch by returning before any hooks are registered or the Agent Control channel is started, closing the unenforced disable setting flagged in #857. Adds regression tests for the false/true/absent cases and the `tool_approval_enabled` independence, updates README and CHANGELOG, and bumps the plugin to 0.3.1 across its manifests.
>
> <sup>Written by [Preloop PR Reviewer](https://preloop.ai) for commit 421cc3a0. Updates automatically on new commits.</sup>


d-mo:
Thanks for the review on `421cc3a0`. Addressing the two LOW items:

1. **CLI `run` / exported `plugin` singleton.** Agreed that those paths do not consult `enabled`. This slice is the host `register()` path that ClawHub flagged: that is what kept a live control channel after `enabled: false`. `register()` already constructs its own instance and never starts the exported singleton, so a disabled registration still leaves no socket. `preloop-openclaw-plugin run` is an explicit CLI start. I documented that in the README on `e9470583` rather than changing CLI semantics here.

2. **`definition.version` still `0.1.1`.** Leaving it. `PUBLISHING.md` lockstep for this release is `package.json`, `package-lock.json`, and `openclaw.plugin.json` (all 0.3.1). The host reads the manifest, not `definition.version`, and the constant was already stale before this PR. Bumping or dropping it is a follow-up, not part of honouring `config.enabled`.

No inline threads to resolve. Happy to take a follow-up on `definition.version` separately.

